PR #1, opened 2017-09-15
Implement the titan lexer and parser

gligneul:
There are few thing pending, but I think we should create new issues to fix them.

- Lexer:
  + accept multi-line comments/strings
  + hexadecimal numbers
  + full float-point numbers
  + string escapes

- Parser:
  + operator precedence
  + add locations to ast
  + add void type??

- Automatic tests
  + with busted?
  + should we use the inspect library or a custom serializer?

### DIFF
--- a/examples/add.titan
+++ b/examples/add.titan
@@ -1,0 +1,6 @@
+-- Add two numbers
+
+function foo(x: integer, y: integer): integer
+    return x + y
+end
+
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -1,0 +1,80 @@
+local ast = {}
+
+-- Declarations of union types
+local types = {}
+
+types.Type = {
+    Basic   = {'name'},
+    Array   = {'subtype'},
+}
+
+types.TopLevel = {
+    Func    = {'islocal', 'name', 'params', 'rettype', 'block'},
+    Var     = {'islocal', 'decl', 'value'},
+}
+
+types.Decl = {
+    Decl    = {'name', 'type'},
+}
+
+types.Stat = {
+    Block   = {'stats'},
+    While   = {'exp', 'block'},
+    Repeat  = {'block', 'exp'},
+    If      = {'thens', 'elsestat'},
+    For     = {'name', 'start', 'end', 'inc', 'block'},
+    Assign  = {'var', 'exp'},
+    Decl    = {'decl'},
+    Call    = {'callexp'},
+    Return  = {'exp'},
+}
+
+types.Then = {
+    Then    = {'exp', 'block'},
+}
+
+types.Var = {
+    Name    = {'name'},
+    Index   = {'exp', 'exp'},
+}
+
+types.Exp = {
+    Nil     = {},
+    Bool    = {'bool'},
+    Integer = {'integer'},
+    Float   = {'float'},
+    String  = {'string'},
+    Table   = {'exps'},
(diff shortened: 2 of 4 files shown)


mascarenhas (approved):
It is a good start, we have to check with Sérgio how to do left-associativity with parser-gen, I didn't see anything like chainl or fold captures in the docs. At the worst we can fake it with repetition, throwing all the expressions with the same precedence level plus their operators in a table, and do a left fold on that table.
By the way, I do not think v0 needs an explicit void type, functions that do not return anything can be assumed to return nil.

andremm:
> It is a good start, we have to check with Sérgio how to do left-associativity with parser-gen, I didn't see anything like chainl or fold captures in the docs. 

@mascarenhas maybe `parser-gen` does not have `chainl` or fold captures because it is based on the `re` module from LPeg, which also does not provide fold captures.

>At the worst we can fake it with repetition, throwing all the expressions with the same precedence level plus their operators in a table, and do a left fold on that table.

For now I believe we will have to go for this kind of workaround. I'm posting here a silly example that shows how we can implement it in the Titan parser. :)

```lua
package.path = package.path .. ";../?.lua"
local pg = require "parser-gen"

local function eval (exp)
  if exp.rule == "exp" then
    if exp[2] then
      return eval(exp[1]) + eval(exp[3])
    else
      return eval(exp[1])
    end
  elseif exp.rule == "term" then
    if exp[2] then
      return eval(exp[1]) * eval(exp[3])
    else
      return eval(exp[1])
    end
  elseif exp.rule == "factor" then
    if exp[2] then
      return - eval(exp[2])
    else
      return eval(exp[1])
    end
  elseif exp.rule == "power" then
    if exp[2] then
      return eval(exp[1]) ^ eval(exp[3])
    else
      return eval(exp[1])
    end
  elseif exp.rule == "primary" then
    return eval(exp[1])
  else
    return tonumber(exp[1])
  end
end

local function previous_rule (rule)
  if rule == "term" then
    return "exp"
  else
    return "term"
  end
end

local function fold (...)
  local t = { ... }
  local rule = previous_rule(t[1].rule)
  local folded = { rule = rule, t[1] }
  for i=2,#t,2 do
    folded = { rule = rule, folded, t[i], t[i+1] }
  end
  return folded
end

local grammar = pg.compile([==[
  exp <- (term ('+' term)*) -> fold !.
  term <- (factor ('*' factor)*) -> fold
  factor <- '-' factor / power
  power <- primary ('^' factor)?
  primary <- NUM
  NUM <- [0-9]+
]==], { fold = fold })

local ast, err = pg.parse("-1+2^3+2*5", grammar)
if ast then
  print(eval(ast))
else
  print("error")
end
```

hugomg:
The solution I am coding right now is to dump all the basic expressions
and operators on a flat list and use a smart precedence-aware algorithm
to build a tree after the fact. I is similar to the one used in the
subexpr function in Lua's lparser.c, but with a different logic for the
unary prefix operators.

One thing that I am finding a bit tricky is that it is very easy to mix
`<` with `<<`, `local x` with `localx`, and so on. In this particular
case of parsing this luckly doesn't bite us because none of the prefix
operators is the suffix of another operator but I find this very
fragile... I am thinking about moving all the quoted strings in the
parser section of the grammar to the lexer section and use lookaheads
to make sure we never match a prefix of a longer token. The kind of
thing that the symb and kw functions did in TypedLua's lexer.

PR #8, opened 2017-09-18
Lexer

hugomg:
I split the lexer into a separate module and wrote a test suite for it with busted. The big change here is that the lexer obeys the "longest match" rule: for example, the pattern for "<" won't match "<=" and the pattern for "local" won't match "localx = 10". This is done adding appropriate negative lookaheads to the patterns.

I still need to figure out how to report lexing (and parsing) errors. I would like have the lexer report errors for invalid numbers, unclosed strings, and so on without silently backtracking but I don't know what is the best way to do it. Raise a Lua `error`? Use lpeglabel error labels? Something else?

----

I also have a few other questions:

* if we go for the `error` route, how can I catch the lexer and parser errors and reraise any other errors? I forgot if there is a way to do this with pcall.

* Last time we closed the pull request with a "merge". Perhaps we should use "rebase and merge" to keep a cleaner history in the master branch?

### DIFF
--- a/spec/lexer_spec.lua
+++ b/spec/lexer_spec.lua
@@ -1,0 +1,240 @@
+local lexer = require 'titan-compiler.lexer'
+local lpeg = require 'lpeglabel'
+
+local function table_extend(t1, t2)
+    table.move(t2, 1, #t2, #t1+1, t1)
+end
+
+-- Find out how the lexer lexes a given string.
+-- Produces an error message if the string cannot be lexed or if there
+-- are multiple ways to lex it.
+local function run_lexer(source)
+    local all_matched_tokens = {}
+    local all_captures = {}
+
+    local i = 1
+    while i <= #source do
+
+        local found_token = nil
+        local found_captures = nil
+        local found_j = nil
+
+        for tokname, tokpat in pairs(lexer) do
+            local captures, j = lpeg.match(lpeg.Ct(tokpat) * lpeg.Cp(), source, i)
+            if captures then
+                if i == j then
+                    return string.format(
+                        "error: token %s matched the empty string",
+                        tokname)
+                elseif found_token then
+                    return string.format(
+                        "error: multiple matching tokens: %s %s",
+                         found_token, tokname)
+                else
+                    found_token = tokname
+                    found_captures = captures
+                    found_j = j
+                end
+            end
+        end
+
+        if not found_token then
+            return "error: lexer got stuck"
+        end
+
+        table.insert(all_matched_tokens, found_token)
+        table_extend(all_captures, found_captures)
+        i = found_j
+    end
+
+    return { tokens = all_matched_tokens, captures = all_captures }
+end
+
+local function assert_lex(source, expected_tokens, expected_captures)
+    local lexed    = run_lexer(source)
+    local expected = { tokens = expected_tokens, captures = expected_captures }
+    assert.are.same(expected, lexed)
+end
+
+describe("Titan lexer", function()
+
+    it("can lex some keywords", function()
+        assert_lex("and", {"AND"}, {})
+    end)
+
+    it("can lex keywords that contain other keywords", function()
+        assert_lex("if",     {"IF"},     {})
+        assert_lex("else",   {"ELSE"},   {})
+        assert_lex("elseif", {"ELSEIF"}, {})
+    end)
+
+    it("does not generate semantic values for keywords", function()
+        assert_lex("nil",   {"NIL"},   {})
+        assert_lex("true",  {"TRUE"},  {})
+        assert_lex("false", {"FALSE"}, {})
+    end)
+
+    it("can lex some identifiers", function()
+        assert_lex("hello",    {"NAME"}, {"hello"})
+        assert_lex("_",        {"NAME"}, {"_"})
+        assert_lex("_test_17", {"NAME"}, {"_test_17"})
+    end)
+
+    it("can lex identifiers containing keywords", function()
+        assert_lex("return17", {"NAME"}, {"return17"})
+        assert_lex("andy",     {"NAME"}, {"andy"})
+        assert_lex("_end",     {"NAME"}, {"_end"})
+    end)
+
+    it("can lex sequences of symbols without spaces", function()
+        assert_lex("+++", {"ADD", "ADD", "ADD"}, {})
+        assert_lex(".",   {"DOT"},               {})
+        assert_lex("..",  {"CONCAT"},            {})
+        assert_lex("...", {"DOTS"},              {})
+        assert_lex("<==", {"LE", "ASSIGN"},      {})
+        assert_lex("---", {"COMMENT"},           {})
+        assert_lex("///", {"IDIV", "DIV"},       {})
+        assert_lex("~=",  {"NE"},                {})
+        assert_lex("~~=", {"BXOR", "NE"},        {})
+    end)
+    
+    it("can lex some integers", function()
+        assert_lex("0",            {"NUMBER"}, {0})
+        assert_lex("17",           {"NUMBER"}, {17})
+        assert_lex("0x1abcdef",    {"NUMBER"}, {0x1abcdef})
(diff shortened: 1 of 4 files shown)


titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+-- The exported tokens are in ALLCAPS, to play nice with the "re"-style grammars
+-- from parser-gen: it can only refer to our tokens if we use alphabetical
+-- identifiers and it expects terminals to be uppercase and non-terminals to be
+-- lowercase.
+
+local lpeg = require 'lpeglabel'
+local P = lpeg.P
+local R = lpeg.R
+local S = lpeg.S
+local C = lpeg.C
```
gligneul: This could be written in one line.

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+-- weird things such as `1337require`.
+local number_start = P(".")^-1 * R("09")
+local expo = S("EePp") * S("+-")^-1
+local possible_number = #number_start * (expo + R("09", "AZ", "az") + ".")^1
+lexer.NUMBER = Cmt(possible_number, function(_, i, s)
+    local n = tonumber(s)
+    if n then
+        return i, n
+    else
+        assert(false, "todo: nice syntax error for malformed number")
```
gligneul: remove the if and insert the assert before: `assert(n, ...`

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+end)
+
+--
+-- Strings
+--
+
+-- if s[i] points to a newline sequence, skip it
+local function skip_linebreak(s, i)
+    local c1, c2 = string.byte(s, i, i+1)
+    if     c1 == 10 and c2 == 13 then return i + 2
```
gligneul: We need to add a comment stating what is '10' and '13'.
Also, why do we need \r\n and \n\r?

gligneul: Could this be a if, elseif with two cases, and without the fancy alignment?

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+
+local longstring_open  = P("[") * C( P("=")^0 ) * P("[")
+local longstring = Cmt(longstring_open, function(source, i, equals)
+    i = skip_linebreak(source, i)
+    local close_str = "]" .. equals .. "]"
+    local j, k = string.find(source, close_str, i, true)
+    if j then
+        return k+1, string.sub(source, i, j-1)
+    else
+        assert(false, "TODO: friendly syntax error for EOF")
```
gligneul: replace assert(false,

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+    end
+end)
+
+local escapes = {
```
gligneul: Are all escapes here?

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+    ["r"]  = "\r",
+    ["t"]  = "\t",
+    ["v"]  = "\v",
+    ["\\"] = "\\",
+    ["\'"] = "\'",
+    ["\""] = "\"",
+}
+
+local shortstring = Cmt( C(P('"') + P("'")), function(source, i, delimiter)
+    local parts = {}
```
gligneul: This is the output string characters, so why not chars instead of parts?

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+
+    while true do
```
gligneul: I would like to see this while true with something like `while i < #str then.
Then, in the loop body we should process each char individually.

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+
+        table.insert(parts, s)
+
+        if not c then
+            assert(false, "TODO: friendly syntax error for unfinished string (EOF)")
+        elseif c == "\n" or c == "\r" then
+            assert(false, "TODO: friendly syntax error for unfinished string (\\n)")
+        elseif c == delimiter then
+            break
+        elseif c == "\\" then
```
gligneul: We should create a function to process the escapes, the loop is too big and make lot of different things.

spec/lexer_spec.lua
```diff
@@ -0,0 +1,224 @@
...
+
+    it("can lex some short comments", function()
+        assert_lex("if--then\nelse", {"IF", "COMMENT", "ELSE"}, {})
+    end)
+
+    it("can lex short comments that go until the end of the file", function()
+        assert_lex("--aaaa", {"COMMENT"}, {})
+    end)
+
+    it("can lex long some comments", function()
```
gligneul, line 215: We could add this test:
```
---[[
print('hello')
--]]
```

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+local short_comment = comment_start * (P(1) - P("\n"))^0 * P("\n")^-1
+local long_comment  = comment_start * longstring / function() end
+
+lexer.COMMENT = long_comment + short_comment
+
+--
+-- Keywords and names
+--
+
+local idstart = R("AZ", "az", "__")
```
gligneul: This could be:
```
local idstart =  P("_") + R("AZ", "az")
local idrest = idstar + R("09")
```

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+local idrest  = R("AZ", "az", "__", "09")
+local possiblename = idstart * idrest^0
+
+local keywords = {
+    -- Lua keywords:
```
gligneul: I don't think this commentary is necessary yet.

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+    NE = "~=", LE = "<=", GE = ">=",
+    ASSIGN = "=",
+    LPAREN   = "(", RPAREN   = ")",
+    LBRACKET = "[", RBRACKET = "]",
+    LCURLY   = "{", RCURLY   = "}",
+    SEMICOLON = ";", COMMA = ",", 
+    DOT = ".", DOTS = "...", DBLCOLON = "::", 
+    -- Titan:
+    COLON = ":",
+
```
gligneul: Remove this `\n`

titan-compiler/lexer.lua
```diff
@@ -0,0 +1,217 @@
...
+}
+
+for tokname, symbol in pairs(symbols) do
+    local pat = P(symbol)
+    for _ , symbol2 in pairs(symbols) do
+        if #symbol < #symbol2 and symbol == string.sub(symbol2, 1, #symbol) then
```
gligneul, line 209: Could we just use string.find?

titan-compiler/parser.lua
```diff
@@ -199,7 +199,7 @@ local function errorstostr(errors)
 end
 
 function parser.parse(input)
-	local result, errors = pg.parse(input, grammar)
+    local result, errors = pg.parse(input, grammar)
```
gligneul, line 202: I wouldn't commit this on this PR.

spec/lexer_spec.lua
```diff
@@ -0,0 +1,224 @@
...
+    end)
+    
+    it("can lex some integers", function()
+        assert_lex("0",            {"NUMBER"}, {0})
+        assert_lex("17",           {"NUMBER"}, {17})
+        assert_lex("0x1abcdef",    {"NUMBER"}, {0x1abcdef})
+        assert_lex("0x1ABCDEF",    {"NUMBER"}, {0x1ABCDEF})
+    end)
+
+    it("can lex some floats", function()
```
gligneul, line 108: TODO: test hexa floats

gligneul (changes requested):

gligneul (approved):

PR #10, opened 2017-09-19
Add proper exception handling

gligneul:
Now we are able to use `error/assert` to internal errors and `exception.throw` to expected errors.
PS: this is just the basic idea, we can implement a catch/finally latter.

### DIFF
--- a/spec/exception_spec.lua
+++ b/spec/exception_spec.lua
@@ -1,0 +1,40 @@
+local exception = require 'titan-compiler.exception'
+
+describe("Titan exception", function()
+
+    it("can print an exception", function()
+        local ex = exception.new("MyException", "error message")
+        assert.are.same(tostring(ex), "error message")
+    end)
+
+    it("can try", function()
+        local ok, err = exception.try(function()
+            return "magic"
+        end)
+        assert.are.same(ok, true)
+        assert.are.same(err, "magic")
+    end)
+
+    it("can throw an exception", function()
+        local ok, err = exception.try(function()
+            exception.throw("MyException", "error message")
+        end)
+        assert.are.same(ok, false)
+        assert.are.same(tostring(err), "error message")
+    end)
+
+    it("can rethrow an exception", function()
+        local ok, err = exception.try(function()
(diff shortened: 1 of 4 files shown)


hugomg:
I would like to see some extra tests clarifying how regular Lua exceptions interact with this module. 
* What happens if we call `error` inside a `try` block?
* What happens if we call `rethrow` with an exception originally raised with `error`?
* What happens if we call `throw` inside a regular `pcall`?
* What happens if we call `throw` inside an unprotected function? Does the programmer receive a good stack trace including source file and line number?

I also wonder if we what we are doing is idiomatic Lua.

hishamhm:
> I also wonder if we what we are doing is idiomatic Lua.

It's always tempting to implement in Lua features like that (heck, at one point I even gave `switch/case` a shot) but it's the kind of thing you end up dropping after using for a while. There's also a whole philosophical discussion as to what is really "exceptional" behavior, and if the control flow of "expected errors" should be made implicit.

The typical use I see in Lua is to just use the `nil, err` protocol for forwarding errors. It is cumbersome to write but easy to follow. Myself, I only use `pcall` to handle things that are "beyond my control" (e.g. calling third-party code with `require`, etc.) I believe modules should only `error` if used incorrectly by the programmer (e.g. type assertion for arguments); values coming from the user should never trigger `error`.

For a compiler, in particular, I don't see a big improvement with an implicit exception system, because:
* user errors are all caught by the front end
  * of those, the lexer and parser are pretty "systematic" in how they report errors already
  * the type checker is one thing you'll eventually want to report multiple errors, and not throw an exception when the first one happens (the alternative would be to start catching in all places you could recover, and then your flow is verbose again but only "kind of explicit")
* in the back end the only kind of errors are internal assertions

At this point, I think it would be best to apply the <a href="https://en.wikipedia.org/wiki/You_aren't_gonna_need_it">YAGNI</a> principle here.

gligneul:
Thanks for the thoughts guys, I'm going to stick with the `nil, err` protocol for now and cancel this PR.

PR #11, opened 2017-09-20
Parser improvements

### LINKED ISSUE #5: Parser should support operator precedence

hugomg:
- The new lexer to correctly parses things like `localx` and shouldn't get confused between `<`, `<=`, `<<` and so on.
- The parser now has proper operator precedence (Fixes #5)

Still missing:
- Test syntax errors (should we switch to PEG error labels first?)
- Add source code positions to the generated AST (so the typechecker can report errors too)


spec/parser_spec.lua
```diff
@@ -0,0 +1,79 @@
+local parser = require 'titan-compiler.parser'
+
+local function slurp(filename)
```
gligneul: We call this function `get_contents` in `titanc`. We should have only one copy of it.

spec/parser_spec.lua
```diff
@@ -0,0 +1,79 @@
...
+
+local function assert_parse(sourcefilename)
+    local source = slurp(sourcefilename)
+    local expected = slurp(sourcefilename..".ast")
+    local ast, errors = parser.parse(source)
+    local obtained
+    if ast then
+        obtained = parser.pretty_print_ast(ast).."\n"
+    else
+        error("not yet implemented")
```
gligneul, line 12: NYI? 😱

titanc
```diff
@@ -35,6 +35,6 @@ else
 end
 local ast = check(parser.parse(input))
 if args.print_ast then
-    print(inspect(ast))
+    print(parser.pretty_print_ast(ast))
```
gligneul, line 31: Remember to remove the `require 'inspect'` in the top of the file.

titan-compiler/parser.lua
```diff
@@ -21,170 +47,266 @@ function defs.boolopt(x)
...
-        elseif suffix._typename == 'Exp' then
-            local var = ast.Var.Index(exp, suffix)
-            exp = ast.Exp.Var(var)
-        else
-            error('impossible')
-        end
+function defs.fold_suffixes(exp, suffixes)
+    for i = 1, #suffixes do
+        local suf = suffixes[i]
+        exp = suf(exp)
```
gligneul: I think we need at least a commentary explaining that suffixes captures return functions.

gligneul: We may also rename `suffix` to `applysuffix`

titan-compiler/parser.lua
```diff
@@ -21,170 +47,266 @@ function defs.boolopt(x)
...
-                     / (tablecons)                      -- produces Exp
-                     / (functioncall)                   -- produces Exp
-                     / (var)                            -> Exp_Var
-                     / ('(' exp ')')                    -- produces Exp
+    decl            <- (NAME (COLON type)? -> opt)      -> Decl_Decl
 
-    exp             <- (unop exp)                       -> Exp_Unop
-                     / (value binop exp)                -> Exp_Binop
-                     / (value)                          -- produces Exp
+    type            <- (NIL->'nil')                     -> Type_Basic
```
gligneul: replace `NIL->'nil'` with `NIL -> 'nil'`

titan-compiler/parser.lua
```diff
@@ -21,170 +47,266 @@ function defs.boolopt(x)
...
-
-    COMMENT         <- ('--' (!%nl .)* %nl)
-
-    SKIP            <- (%s / %nl / COMMENT)
+    fieldsep        <- SEMICOLON / COMMA
+
+    -- Create new rules for all our tokens, so parser-gen can
+    -- work its whitespace-skipping magic
+    
+    AND             <- %AND
```
gligneul, line 246: Those could be automatically generated.

gligneul (changes requested):

titan-compiler/parser.lua
andremm: I would use more meaningful names to operations. For instance, instead of 
```
e1              <- {| e2  (op1  e2 )* |}
```
I would use something like
```
e1              <- {| e2  (opOR  e2 )* |}
```
but this is just a suggestion, not a requirement for accepting the PR. :)

gligneul (approved):

PR #12, opened 2017-09-21
Add the variable binding step in the checker module

gligneul:
Also:
- made some improvements on the `util.get_file_contents` file;
- simplified the ast by removing the `_type` field;
- created a method to traverse the ast.

PS: I think we should either remove the `testfiles/*.ast` from the repository or replace them with a format that doesn't break every time we change a little thing in the parser. Any suggestions?


spec/symtab_spec.lua
```diff
@@ -0,0 +1,22 @@
+local symtab = require 'titan-compiler.symtab'
```
hugomg, line 1: I'm not sure we really need this test unless there are some tricky symbol table bugs that you are worrying about. But I guess it doesn't hurt...

spec/symtab_spec.lua
```diff
@@ -0,0 +1,22 @@
...
+
+describe("Titan symbol table", function()
+
+    it("can find some symbols", function()
+        local st = symtab.new()
+        local d_a, d_a2, d_b = {}, {}, {}
```
hugomg: IIRC, assert.are.same does a structural comparision so it can't tell empty tables apart from each other. You could change them to sequential numbers or unique strings.

spec/symtab_spec.lua
```diff
@@ -0,0 +1,22 @@
...
+        st:open_block()
+        st:add_symbol("a", d_a)
+        assert.are.same(st:find_symbol("a"), d_a)
+            st:open_block()
```
hugomg: The indentation here is inconsistent. I would indent after an open_block() and unindent on close_block().

titan-compiler/ast.lua
```diff
@@ -54,21 +54,32 @@ types.Args = {
```
hugomg, line 54: Have you considered storing the node contents in the array part of the table and using the for setindex and getindex?

Pros:
* Can define traverse elsewhere
* Inspect will also print stuff in order

Cons:
* Might be slower
* inspect output migth be less readable.

titan-compiler/util.lua
```diff
@@ -1,10 +1,17 @@
...
+    local f, err = io.open(filename, "r")
+    if not f then
+        return false, err
+    end
     local s = f:read("a")
     f:close()
-    return s
+    if not s then
+        return false, "file '" .. filename .. "' is empty"
+    else
```
hugomg: Do we really need to produce an error on empty files?

titan-compiler/symtab.lua
```diff
@@ -0,0 +1,35 @@
...
+
+function symtab:open_block()
+    table.insert(self.blocks, {})
+end
+
+function symtab:close_block()
+    assert(#self.blocks > 0, "stack underflow")
+    table.remove(self.blocks)
+end
+
```
hugomg, line 17: If you ever have trouble with mismatched open_block and close_block calls, consider creating a symtab:with_block function:

```lua
function symtab:with_block(body)
    self:open_block()
    body()
    self:close_block()
end
```

(Don't need to add it now though, YAGNI and so on).

titan-compiler/ast.lua
```diff
@@ -54,21 +54,32 @@ types.Args = {
 
 -- Create a function for each type constructor
 for typename, conss in pairs(types) do
-    ast[typename] = {}
     for consname, fields in pairs(conss) do
-        ast[typename][consname] = function(...)
+        local tag = typename .. '_' .. consname
+
+        local mt = { __index = {
+            traverse = function(self, visitor, ...)
```
hugomg: This version of traverse is not recursive. Looks more like a foreach / map than a traverse...

titan-compiler/checker.lua
```diff
@@ -0,0 +1,55 @@
+local checker = {}
+
```
hugomg, line 2: I like how short this file got and how we avoided most of the boilerplate. But I suspect that we can improve the traverse function:

* Traverse should be a standalone function instead of a method in the ast nodes. The method on the nodes should be as simple/generic as possible. (When I am doing things like this in Ocaml the nodes have an "fmap" method and then I have separate functions for iteration, folding, etc).

* The traverse function should receive just the node as its parameters. Additional parameters such as `st` should be passed as upvalues.

* The default case of "if none of the above continue iterating" should be built-in to the traverse function.

* The constant "if not ok then return false, err" blocks add a lot of clutter. Traverse should be able to handle this somehow.

titan-compiler/checker.lua
```diff
@@ -0,0 +1,55 @@
...
+        local ok, err = node:traverse(bindvars, st)
+        if not ok then return false, err end
+        st:close_block()
+
+    elseif tag == "Var_Name" then
+        node.decl = st:find_symbol(node.name)
+        if not node.decl then
+            -- XXX generate better error messages when we have the line num
+            return false, "variable '" .. node.name .. "' not declared"
+        end
```
hugomg, line 38: I think that we could collect these errors in a list instead of bailing out after the first one. Undefined variables is not like syntax errors, where having an error at the start leads to tons of nonsense afterwards.

hugomg:
Regarding the AST tests, I think *some* kind of test there is important because that stuff is easy to get wrong (I fixed lots of bugs while building that test suite).

But I agree that changes to those files do clutter the diffs a lot. I wonder if we could use a more stable serialization format that is less prone to changing all the time. Alternatively, we could restrict the tests to only look at one portion of the AST, so they won't be affected by unrelated changes in the AST. One way to do it would be to manually include assertions inside the spec file that look at just the parts of the ast we are interested in for each test, instead of using the reference ".ast" files. Another way would be to include just a subset of the tree in the .ast file and test it with an "assert is subset" test instead of an "assert is same" one.

titan-compiler/checker.lua
hugomg, line 2: I think I would prefer something similar to the visitor API from [this Typed Lua PR](https://github.com/jvprat/typedlua/blob/0569706b0a68a9adfe9aec1ff7146af619008905/typedlua/tlvisitor.lua). Just add an additional "visit_children" method and define the default traversal using pairs/ipairs instead of the dozens of mutually recursive methods.

spec/symtab_spec.lua
hishamhm: assert.equal checks for plain equality, but I agree unique strings would be clearer.

titan-compiler/symtab.lua
hishamhm: I love this style of wrapper functions because it states intent in the code so clearly. Can't wait until the day comes and we have aggressive inlining in Titan so I can use this freely without Lua people nagging me with "but you're creating a closure every time!"

hishamhm:
> One way to do it would be to manually include assertions inside the spec file that look at just the parts of the ast we are interested in for each test, instead of using the reference ".ast" files.

Yes, I would go with that approach. If I want to test that a node in a nested table is present, I just do `assert.string(my.nested.table[2].bla)` (if the traversal crashes, that's just a failed test). That way we check for what the test is testing, and if something else in the tree is relevant, it should be a separate assert (or even a separate test).

titan-compiler/ast.lua
hishamhm: > inspect output might be less readable.

This is a major con. Not only for inspect, but for the rest of the code that uses the AST. See the Typed Lua codebase, full of stuff like `stat[1][3][2][1].type == stat[1][3][1].type`. It takes someone deeply entrenched in that codebase like @andremm was during his PhD to be able to glance that and understand what does it mean (and I doubt he still remembers the meaning of all those numbers there as quickly today!).

spec/symtab_spec.lua
hishamhm: Having gone through the whole PR, I think this is the only blocker. Some of @hugomg's suggestions can always be integrated later but this PR move things forward rather nicely!

hishamhm (changes requested):

titan-compiler/ast.lua
andremm: > See the Typed Lua codebase, full of stuff like stat[1][3][2][1].type == stat[1][3][1].type

@hishamhm I'm not sure where did you find this kind of construction. I just ran `grep -r "stm\[" * | grep "\.type"`  in the Typed Lua code base and couldn't find them. :)
I used `grep` with `stm` instead of `stat` because it is `stm` that I use to access statement nodes instead of using `stat`.

Anyway, I agree that inspect output being less readable is a major con.

andremm (changes requested):

andremm (approved):

andremm (commented):
It is getting great! In further PRs I would change
```lua
local types = {}
```
to
```lua
local nodes = {}
```
in `ast.lua` because calling AST nodes as AST types is a bit odd to me. :)

hugomg (approved):

titan-compiler/ast.lua
hishamhm: @andremm I was paraphrasing :) And the code has changed a lot since I last played with it! I had to dig back in to the Git history, I was thinking of stuff like this: https://github.com/andremm/typedlua/blob/9b9bd5d1ba4f8069830302fefb7ff91cf3ba2343/typedlua/tlchecker.lua#L1156-L1160 (this is the most pathological example I ran into, but there's lots of stuff like `[2][1]` out there which assume you have the AST format in your head.)

andremm: hahahaha, yeah, you have a very good point @hishamhm. The first versions of the code that implemented the filters of union types inside if statements were kind of convoluted and assumed you had the AST stuck in your mind. :)

titan-compiler/util.lua
gligneul: If we don't produce an error, i think we should at least return ''.
What do you think?

titan-compiler/ast.lua
gligneul: I'm going to rename it to `foreach`.

titan-compiler/util.lua
hugomg: Yes, I think returning the empty file would be the best case here. I think read("a") already does that though, without needing to treat it separately.

gligneul: That's true, but on my system `f:read` returns nil if the file is a directory.
`lua -e "f = io.open('.'); print(type(f:read('a')))"`
I think we should return a generic error message in this case.

spec/symtab_spec.lua
gligneul: fixed

titan-compiler/symtab.lua
gligneul: I like the idea, but this make the error treatment awkward in the check module:
```
local ok, err = st:with_blocks(function() 
    local ok, err = node:foreach(bindvars, st)
    if not ok then return false, err end
end)
if not ok then return false, err end
```

gligneul: In fact, I'm going to remove the `if not ok ...` and implement this.

titan-compiler/checker.lua
gligneul: We can have a clear API if we just implement the __pairs metamethod.

titan-compiler/checker.lua
gligneul: Agreed.

hishamhm (approved):

PR #13, opened 2017-09-23
Improve the ast iteration and checker error handling

gligneul:
- Replace the foreach method with an iterator.
- Remove the `return false, err` from bindvars function, now it returns
  a table with the errors
- Replace the symtab functions {open/close}_block with with_blocks.


titan-compiler/ast.lua
```diff
@@ -57,15 +57,17 @@ for typename, conss in pairs(types) do
     for consname, fields in pairs(conss) do
         local tag = typename .. '_' .. consname
 
+        local function iter(node, i)
+            i = i + 1
+            if i <= #fields then
+                return i, node[fields[i]]
```
hugomg, line 63: This should return `fields[i], node[fields[i]]` instead of `i, node[fields[i]]`. The `i` doesn't have any meaning for those calling this function.

titan-compiler/symtab.lua
```diff
@@ -6,13 +6,18 @@ function symtab.new()
...
 end
 
-function symtab:close_block()
-    assert(#self.blocks > 0, "stack underflow")
-    table.remove(self.blocks)
+local function close_block(st)
+    table.remove(st.blocks)
+end
+
+function symtab:with_blocks(body, ...)
```
hugomg: Shouldn't this be `with_block`, in singular?

titan-compiler/symtab.lua
```diff
@@ -6,13 +6,18 @@ function symtab.new()
     return setmetatable({ blocks = {} }, symtab)
 end
 
-function symtab:open_block()
-    table.insert(self.blocks, {})
+local function open_block(st)
```
hugomg: We might want to leave `open_block` and `close_block` as public functions, since they form a more flexible API.

hugomg (changes requested):
Looks great overall. I like the idea of the children iterator.

titan-compiler/ast.lua
gligneul: I agree, but it is possible to make a stateless iterator if I return `fields[i]` instead of `i`?

titan-compiler/symtab.lua
gligneul: Agreed, fixed.

titan-compiler/symtab.lua
gligneul: Fixed.

titan-compiler/ast.lua
hugomg: Hmm... It would be possible if we kept an index somewhere mapping each key to the next. But that might not be worth the extra complexity.

hugomg (approved):

PR #14, opened 2017-09-24
Get rid of ".ast" files in parser tests

hugomg:
Now the tests for the parser only look at the part of the AST related to
the feature they are testing and don't care if the structure of an
unrelated part changes.

This should hopefuly reduce the number of tests that break when we
update the AST, as well as the clutter in the patch diffs.


hishamhm:
LGTM!

A possible idea for the future would be to inline the small .titan unit testcases directly in the spec file; reading the testcase and the AST fragment side-by-side would make the tests even easier to follow. (A valid argument against that would be if they are to be reused when testing the next phases of the compiler.)

hugomg:
> A possible idea for the future would be to inline the small .titan unit testcases directly in the spec file;

I don't think it is likely that these tests will be reused in later steps. Most of them won't type check.

That said, one advantage of the tests being in a separate file is that we can pass them to titanc --print-ast to see how Titan is parsing them. At the moment this is the only way to debug a failing test case because assert.are.same's error message only shows the first 3 levels or so of the mismatched AST.

gligneul (approved):
Nice job, way better than the `.ast` files.

PR #16, opened 2017-09-26
Improvements in AST and checker module

gligneul:
- Rename fields in the AST.
- Add test case for `for` statement.
- Create a generic function to visit the children in the checker module.


testfiles/statements.titan
```diff
@@ -14,5 +14,9 @@ local function statements(): nil
         print("Hello", "World")
     end
 
+    for i = 1, 2, 3 do
```
hugomg, line 17: I can't get my mind to accept that this loop iterates only 1 instead of 1 2 3 :)

titan-compiler/checker.lua
```diff
@@ -2,32 +2,35 @@ local checker = {}
...
+        for _, child in ipairs(node) do
+            f(child, ...)
+        end
+    else
+        for _, child in node:children() do
+            f(child, ...)
+        end
+    end
+end
+
```
hugomg, line 17: I think this smells funny. I created a [new issue](https://github.com/titan-lang/titan-v0/issues/17) to discuss if we should change this in a future commit.

EDIT: I was talking about the visit_children function in [line 5](https://github.com/titan-lang/titan-v0/pull/16/files/3aeb8eea093113750dd50c4ad0ad3b503739fc14#diff-b1e6cd54f6850187e36b789847922177R5)

titan-compiler/parser.lua
```diff
@@ -135,7 +135,9 @@ function defs.exp_is_call(_, pos, exp)
     end
 end
 
--- Add libraries to defs
+function defs.name2decl(name)
+    return ast.Decl_Decl(name, false)
+end
```
hugomg: If we change the for loop syntax to allow an optional type annotation then we can get rid of this.

hugomg (changes requested):
Looks good to me, but we should change the for loop syntax to allow optional type declarations like everything else.

andremm (commented):
Any reason why having a generic node Value in the AST instead of specific nodes for nil, false, true, integer, float, and string values? Having specific nodes would make the compiler not depend on type and math.type functions to type check them.

hugomg:
The lexer can't tell apart integers from floats and that is by design. It makes it behave more similar to Lua's and allows us to treat stuff like `4..` and `1337require` as syntax errors.

That said, not that you mention it I think it might be cleaner to do this math.type logic in the parser instead of deferring it to the typechecker.

hugomg (approved):

PR #18, opened 2017-09-26
Split Exp_Value into Exp_Nil, Exp_Bool, etc

hugomg:
This should help the type checker because it won't need to mess with the
type and math.type functions. It should also help the parser tests tell
int and floats apart. Busted uses the == operator under the hood, which
does automatic type coercion (0 == 0.0)

### DIFF
--- a/titan-compiler/parser.lua
+++ b/titan-compiler/parser.lua
@@ -21,10 +21,6 @@ for tag, cons in pairs(ast) do
     defs[tag] = cons
 end
 
-function defs.tonil()
-    return nil
-end
-
 function defs.totrue()
     return true
 end
@@ -43,6 +39,22 @@ end
 
 function defs.boolopt(x)
     return x ~= ''
+end
+
+function defs.nil_exp(--[[ s ]])
+    -- We can't call Exp_Nil directly in the parser because we
+    -- need to drop the string capture that comes by default.
+    return ast.Exp_Nil()
+end
+
+function defs.number_exp(n)
+    if math.type(n) == 'integer' then
+        return ast.Exp_Integer(n)
+    elseif math.type(n) == 'float' then
+        return ast.Exp_Float(n)
+    else
+        error("impossible")
+    end
 end
 
 function defs.name_exp(name)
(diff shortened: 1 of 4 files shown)


titan-compiler/parser.lua
```diff
@@ -45,6 +41,22 @@ function defs.boolopt(x)
...
+function defs.number_exp(n)
+    if math.type(n) == 'integer' then
+        return ast.Exp_Integer(n)
+    elseif math.type(n) == 'float' then
+        return ast.Exp_Float(n)
+    else
+        error("impossible")
+    end
+end
+
```
andremm, line 59: I still believe that we might be able to do this down to the lexical level and remove the dependency of `math.type`. Please, check out this [commit](https://github.com/titan-lang/titan-compiler/commit/efb5934d9489878159176249a1719d4734ae6df6) I made in the old repository.

titan-compiler/parser.lua
```diff
@@ -214,11 +226,11 @@ local grammar = pg.compile([[
...
                      / (LBRACKET exp RBRACKET)          -> suffix_index
 
-    simpleexp       <- (NIL -> tonil)                   -> Exp_Value
-                     / (FALSE -> tofalse)               -> Exp_Value
-                     / (TRUE -> totrue)                 -> Exp_Value
-                     / (NUMBER)                         -> Exp_Value
-                     / (STRING)                         -> Exp_Value
+    simpleexp       <- (NIL)                            -> nil_exp
+                     / (FALSE -> tofalse)               -> Exp_Bool
+                     / (TRUE -> totrue)                 -> Exp_Bool
```
andremm, line 231: We may want to split the tag Bool into True and False to comply with our manual:
> The type number is syntax sugar for integer | float, and boolean is syntax sugar for true | false.

andremm (approved):

titan-compiler/parser.lua
hugomg: Doesn't separating things in the lexer make it harder to flag stuff like `4..` and `1337require` as syntax errors (as opposed to a sequence of two tokens)?

titan-compiler/parser.lua
hugomg: Good point but maybe we should leave union types for v1...

PR #19, opened 2017-09-28
Implement parser and syntax errors

hugomg:
I'm not super happy about the current state of this code because I felt like I was fighting against parser-gen's and lpeglabel's APIs to get them to work with each other. In particular:

* The API for setting error labels in those two libraries is incompatible, so it is hard to use an lpeglabel lexer inside a parser-gen parser. If only lpeglabel allowed arbitrary Lua values as labels...
* I'm not sure I really understand how the "%" (subparser) operator works in parser-gen. From the experiments I made, it seemed that the SKIP and error recovery from the subparser can leak into the parent parser, which is weird. The patterns that parser-gen creates also need to be executed with parser-gen.parse and will error-out if you call the match method on them so I am not sure if it is even correct to treat them as patterns.

Anyway, I'm submitting this PR so we can have a place to discuss these problems and because I wanted to get this stuff out of the way so I can can focus on the rest of the compiler for now. I guess we could wait until we figure out a more permanent solution for those lpeg and parser-gen issues before merging this.

### DIFF
--- a/spec/lexer_spec.lua
+++ b/spec/lexer_spec.lua
@@ -1,4 +1,5 @@
 local lexer = require 'titan-compiler.lexer'
+local syntax_errors = require 'titan-compiler.syntax_errors'
 local lpeg = require 'lpeglabel'
 
 local function table_extend(t1, t2)
@@ -20,8 +21,9 @@ local function run_lexer(source)
         local found_j = nil
 
         for tokname, tokpat in pairs(lexer) do
-            local captures, j = lpeg.match(lpeg.Ct(tokpat) * lpeg.Cp(), source, i)
-            if captures then
+            local a, b, c = lpeg.match(lpeg.Ct(tokpat) * lpeg.Cp(), source, i)
+            if a then
+                local captures, j = a, b
                 if i == j then
                     return string.format(
                         "error: token %s matched the empty string",
@@ -35,6 +37,10 @@ local function run_lexer(source)
                     found_captures = captures
                     found_j = j
                 end
+            elseif b and b > 0 then
+                local errnum, suffix = b, c
+                local errcode = syntax_errors.int_to_label[errnum]
+                return { err = errcode }
             end
         end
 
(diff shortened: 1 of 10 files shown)


hishamhm:
I believe @sqmedeiros is the ideal person to give feedback on this. I'm sure he will appreciate seeing the libraries being put through their paces in this realistic scenario.

> Anyway, I'm submitting this PR so we can have a place to discuss these problems and because I wanted to get this stuff out of the way so I can can focus on the rest of the compiler for now. I guess we could wait until we figure out a more permanent solution for those lpeg and parser-gen issues before merging this.

Fully agreed! We can revisit this more calmly (and even propose changes to lpeglabel/parser-gen?) after the workshop. We only need syntax-error-free programs for demoing anyway. :)

mascarenhas (approved):

sqmedeiros:
Hi, everybody.

My main idea when I proposed parser-gen was to make easy to describe a _vanilla_ parser, so a user does not need to worry with some things, such as the matching of whitespaces. At first, I did not imagine that parser-gen would be used together with lpeglabel, so we unfortunately did not pay attention to the integration between both tools.

Fabio already told me that parser-gen has a bug related to error recovery, but I think we are still not dealing with this. But as parser-gen is not mature yet Hugo may have found another bug.

I saw that Hugo open an issue in the lpeglabel repository, but I did not have time this week to look it.

When Fabio added me to Titan, I was planning to write a parser with some error recovery mechanism. As the initial Titan parser was written using parser-gen, my initial schedule was:

1. Fix parser-gen bugs
2. Review the API of LPegLabel and adapt parser-gen
3. Write the Titan parser hopefully using parser-gen

This probably will not be done until the Lua Workshop.
Next week I have more time, so I can help with any issues related to the parser.

hishamhm:
@sqmedeiros Sounds like a great plan! We just need a barebones parser for the workshop prototype. A full-fledged parser with nice error recovery would be a great addition later, and it's something that can be done in parallel as the other guys move on to the type checker and code generator.

PR #20, opened 2017-09-28
first sketch of typechecker

mascarenhas:
This pull request is not meant for merging, just for reviewing and discussion. I have made a first sketch of the typechecker. Major parts missing are checks for unary and binary operations, for integer->float coercions on assignment, for function calls, and for making sure all paths through a function either return or the function has a nil return type.

The typechecker works in two passes, the first one just collects type information for top-level functions and variables (and detects duplicate definitions in the top level), while the second pass does the actual typechecking. All typechecked nodes that have a type get a "_type" field with the type. The types themselves are in "types.lua".

### DIFF
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -24,7 +24,7 @@ types.Stat = {
     If      = {'thens', 'elsestat'},
     For     = {'decl', 'start', 'finish', 'inc', 'block'},
     Assign  = {'var', 'exp'},
-    Decl    = {'decl'},
+    Decl    = {'decl', 'exp'},
     Call    = {'callexp'},
     Return  = {'exp'},
 }
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -49,6 +49,9 @@ types.Exp = {
     Var     = {'var'},
     Unop    = {'op', 'exp'},
     Binop   = {'lhs', 'op', 'rhs'},
+    ToFloat = { 'exp' },
+    ToInt   = { 'exp' },
+    ToStr   = { 'exp' }
 }
 
 types.Args = {
--- a/titan-compiler/checker.lua
+++ b/titan-compiler/checker.lua
@@ -314,7 +314,41 @@ function checkexp(node, st, errors, context)
             error("invalid binary operation " .. op)
         end
     elseif tag == "Exp_Call" then
-        -- TODO: lookup function, check args, _type is return type
+        assert(node.exp._tag == "Var_Name", "function calls are first-order only!")
+        local fname = node.exp.name
+        local func =  st:find_symbol(fname) 
+        if func then
+            local ftype = func._type
+            local nparams = #ftype.params
+            local nargs = #node.args
+            local arity = math.max(nparams, nargs)
+            local moreargs, lessargs
+            for i = 1, arity do
+                local arg = node.arg[i]
+                local ptype = ftype.params[i]
+                if not arg then
+                    atype = ptype
+                else
+                    checkexp(arg, st, errors, ptype)
+                    atype = arg._type
+                end
+                if not ptype then
+                    ptype = atype
+                end
+                checkmatch("argument " .. i .. " of call to function " .. fname, ptype, atype, errors)
+            end
+            if nargs ~= nparams then
+                table.insert(errors, "function " .. fname .. " called with " .. nargs ..
+                    " arguments but expects " .. nparams)
+            end
+            node._type = ftype.ret
+        else
+            table.insert(errors, "function " .. fname .. " not found")
+            for _, arg in ipairs(node.args) do
+                checkexp(arg, st, errors)
+            end
+            node._type = types.Integer
+        end
     else
         error("typechecking not implemented for node type " .. tag)
     end
--- a/titan-compiler/checker.lua
+++ b/titan-compiler/checker.lua
@@ -2,6 +2,7 @@ local checker = {}
 
 local symtab = require 'titan-compiler.symtab'
 local types = require 'titan-compiler.types'
+local ast = require 'titan-compiler.ast'
 
 local checkstat
 local checkexp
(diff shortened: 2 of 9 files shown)


mascarenhas:
Ok, I split the typecheck function into one for statements and one for expressions, the one for statements returns whether that statement "returns" from the current function or not, and this lets us track whether all paths trough the function return something (see `checkfunc`). The `checkexp` function also has a `context` type as a parameter, for now this is only useful when creating empty arrays.

mascarenhas:
Typechecker for binary operations is assuming coercions from integer to floats, and from other types to strings (in concatenation), but still leaving those to be inserted by the code generator. We should probably add explicit coercion nodes, this makes the code generator easier to write.

mascarenhas:
Ok, I added automatic coercions (decorating the AST with new "ToInt", "ToStr", and "ToFloat" nodes) to the typechecker, and it is now complete if it were not untested code full of bugs. Next step is to add the tests and fix the problems that appear.

titan-compiler/symtab.lua
```diff
@@ -11,13 +11,16 @@ function symtab:open_block()
...
 function symtab:close_block()
+    local last = self.blocks[#self.blocks]
     table.remove(self.blocks)
+    return last
 end
 
 function symtab:with_block(body, ...)
     self:open_block()
-    body(...)
+    local res = body(...)
```
gligneul, line 21: I would use `table.pack` here and return with `table.unpack`

titan-compiler/types.lua
```diff
@@ -0,0 +1,56 @@
+
+local types = {}
+
+function types.Function(ptypes, rettype)
```
gligneul, line 4: I think those type definitions could be in `ast.lua` (using the declarative style we have there)

Edit: another option is to put the code that create the constructors from `ast.lua` in another module and use it here.

titan-compiler/types.lua
```diff
@@ -0,0 +1,56 @@
...
+        return types.equals(t1.elem, t2.elem)
+    elseif tag1 == "Function" and tag2 == "Function" then
+        if types.equals(t1.ret, t2.ret) and (#t1.params == #t2.params) then
+            for i = 1, #t1.params do
+                if not types.equals(t1.params[i], t2.params[i]) then
+                    return false
+                end
+            end
+            return true
+        end
```
gligneul, line 39: I would place an `else \n return false \n end` here and remove the `return false` at the end of the function.

hugomg:
Looking at this code more closely, it looks generally good to me (we can always clean things up after the workshop). However, there are still lots of bugs and it currently crashes when we try to typecheck things. Some bugs I noticed:

* In more than one place it does `for i, x in xs` instead of `for i, x in ipairs(xs)`.
* In tests such as `tag == "Array"` it should be `tag == "Type_Array"` instead. Similarly, it is `ast.Exp_toFloat` instead of `ast.toFloat`.

I can try to fix these later tonight. Should I put it in a separate branch or do I push into this branch directly?

mascarenhas:
Please commit directly, the purpose of this branch is to get the typechecker working, so don't worry about pushing broken code. If I need to change anything I will make sure to pull any changes.

titan-compiler/ast.lua
```diff
@@ -49,6 +49,9 @@ types.Exp = {
     Var     = {'var'},
     Unop    = {'op', 'exp'},
     Binop   = {'lhs', 'op', 'rhs'},
+    ToFloat = { 'exp' },
```
gligneul, line 52: Wouldn't it be better to add a `cast` node and have the type as a parameter?

gligneul: Also, there unnecessary extra spaces in this line.

titan-compiler/checker.lua
```diff
@@ -1,57 +1,506 @@
...
-        for _, child in ipairs(node) do
-            f(child, ...)
+local checkstat
+local checkexp
+
+-- Converts an AST type declaration into a typechecker type
+--   typenode: AST node
+--   errors: list of compile-time errors
+--   returns a type (from types.lua)
+local function typefromnode(typenode, errors)
```
gligneul, line 14: Do we need separate structures to represent types?

titan-compiler/checker.lua
```diff
@@ -1,57 +1,506 @@
...
         end
+        return t
+    end
+end
+
+-- Wraps an expression node in a coercion to integer if
+-- type of node is float
+--   node: expression node
+--   returns wrapped node, or original
+local function trytoint(node)
```
gligneul, line 34: I don't like the code repetition here, we should have only a generic function `tryto(type, node)`.

titan-compiler/checker.lua
```diff
@@ -1,57 +1,506 @@
...
+--   annotates the top-level nodes with their types in a "_type" field
+--   annotates whether a top-level declaration is duplicated with a "_ignore" field
+local function firstpass(ast, st, errors)
+    for _, tlnode in ipairs(ast) do
+        local tag = tlnode._tag
+        local name
+        if tag == "TopLevel_Func" then
+            name = tlnode.name
+            local ptypes = {}
+            for _, pdecl in tlnode.params do
```
gligneul: missing ipairs

titan-compiler/checker.lua
```diff
@@ -1,57 +1,506 @@
...
+                table.insert(ptypes, typefromnode(pdecl.type, errors))
+            end
+            tlnode._type = types.Function(ptypes, typefromnode(tlnode.rettype, errors))
+        else
```
gligneul: I prefer `elseif tag == "TopLevel_Var"` here and a `else \n error("impossible")` in the end.

titan-compiler/checker.lua
```diff
@@ -1,57 +1,506 @@
...
+        local msg = "types in %s do not match, expected %s but found %s"
+        table.insert(errors, string.format(msg, term, types.tostring(expected), types.tostring(found)))
     end
+end
 
-    local tag = node._tag
-    if tag == "TopLevel_Func" then
-        st:add_symbol(node.name, node)
-        st:with_block(visit_children, node, st, errors)
+-- Typechecks a repeat/until statement
```
gligneul, line 131: The documentation across the `check*` functions is pretty much the same. 
I think we could have a single one for all of them.

andremm:
>> I can try to fix these later tonight. Should I put it in a separate branch or do I push into this branch directly?

> Please commit directly, the purpose of this branch is to get the typechecker working, so don't worry about pushing broken code. If I need to change anything I will make sure to pull any changes.

@hugomg, please, also do any changes you want in the `codegenerator` branch. Like this branch, the purpose of that branch is to get code generation working.

gligneul (dismissed):
I made several comments about the code, but most of them are small things that can be easily fixed.
I will work on those when I have time.

hugomg:
The latest commit should make the typechecker successfully run for the couple of examples I lifted from the artisanal titan. There likely are still bugs related to the language features that don't appear on those examples -- we will need a more comprehensive test suite to catch those.

hugomg (approved):

titan-compiler/ast.lua
mascarenhas: Technically these are coercions and not casts, as there was still a tiny number of cases I thought making them explicit would simplify the code generator.

titan-compiler/checker.lua
mascarenhas: At first I would have gone with just reusing the AST nodes for the types, but thought that having actual types being separate from the AST nodes used in type declarations would be better separation of concerns, but it works either way.

titan-compiler/checker.lua
mascarenhas: But then these functions would need internally test on `type`, as each one decides differently whether to wrap `node` in a coercion of not, even if we refactor the coercion nodes to be a single node tag parameterized by target type.

titan-compiler/checker.lua
mascarenhas: This would kind of defeat the purpose of having function header comments. :-) The idea is to put these in LDoc syntax or some other comment parser.

gligneul: Couldn't we add a pointer/reference to the first one?
I fear that one of the functions might not be updated and the documentation becomes inconsistent.

titan-compiler/checker.lua
gligneul: I see, makes sense.

titan-compiler/checker.lua
gligneul: We could generate the different versions of the functions with meta-programing.

titan-compiler/ast.lua
gligneul: The generalization might lead to less code repetition elsewhere, but we can improve this later if that is the case.

PR #21, opened 2017-09-29
talking about code generation

andremm:
Following the idea from the PR about sketching the type checker, I'm creating this PR to discuss the code generation. I committed a code generator skeleton for working on code generation in parallel to type checking, that is, the idea is not to merge it right now, but to use this branch for working on code generation at the same time that we use the `typechecker` branch for working on type checking. For now I'm assuming that type checking never fails, so we can generate code for more trivial examples, but at some point we will be able to generate code only if type checking passes.


mascarenhas:
Compilation in a single pass from straight AST that the typechecker gives is probably feasible for a Lua target, but I think compiling to C will be much easier if a first pass converts the typed AST to something like [A-normal form](http://matt.might.net/articles/a-normalization/), where all the operands for compound expressions are either literals, variables, or temporaries.

hugomg:
I think that in the long run we will definitely want an intermediate representation but I suspect that it might be possible to skip that if the goal is to generate something in "artisanal titan" style.

mascarenhas:
The issue I see is when reading from an array inside another expression, reading from an array is a nice chunk of C code, you cannot generate it online with the rest of the code for the expression, you have to lift it to a separate block, store the result in a temporary, and reference it.

Same thing with function calls, as arguments have to be pushed to the Lua stack, and the return value popped. You also cannot do that inline.

With a simple transformation to the AST you could do all this lifting before codegen, making codegen much simpler.

hugomg:
I had some very evil stringly-typed code in mind that would kind of do the A-normal form transformation and the codegen in one step. However, if we can do the transformation without needing to introduce a new datatype for the transformed AST then it is probably saner to do it in two separate steps.

mascarenhas:
There is no need for a new datatype, using Stat_Decl nodes for the temporaries, with generated names, will work fine. The signature for the transformation function takes an expression node and returns the transformed node plus a list of Stat_Decl nodes for the temporaries the transformation introduced.

mascarenhas:
Generating code for Titan->Titan function calls should be straightforward, I will probably have some time to do it before the workshop. Changing the build process to run on stock Lua is trickier, I have to do the transitive closure of the C source files of all the internal Lua functions that the generated code is using, but it would make using the compiler much nicer!

mascarenhas:
One more thing, right now the function prelude is just reserving needed stack space, but not actually incrementing L->top and initializing it, this is done lazily any time a new GC local/temporary is declared. It is probably better to just get a `base pointer`, increment L->top and then `nil` everything in the prelude, so all locals and temporaries can be just `bp + depth`, and messing around with L->top is needed inside the function.

Edit: did this in the next commit. :-)

mascarenhas:
Added a couple of toy, but non-trivial example functions that use tables of tables. Seems to be generating the write barrier all right, but need to think about a way to stress test the GC to make sure this is working.

mascarenhas (approved):
Folks, there are still things missing, and we need a way to script automated tests for this part, but I think we have enough for a demo.

andremm:
Thanks @mascarenhas! That’s great we have a prototype for the workshop!

PR #24, opened 2017-10-06
adding line and column number in type error messages

andremm:
This PR adds the basic code for generating error messages that contain information about line and column number. For instance, when compiling
```lua
local function f():Nil
  return 1
end
```
the current version shows the following error message:
```
./titanc: types in return do not match, expected nil but found integer
```
and the new version will show the following error message:
```
./titanc: t.titan:2:10: type error: types in return do not match, expected nil but found integer
```

### DIFF
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -77,12 +77,12 @@ for typename, conss in pairs(types) do
             end
         }}
 
-        ast[tag] = function(...)
+        ast[tag] = function(pos, ...)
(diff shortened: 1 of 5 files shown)


hishamhm:
I added the binary search to find positions, recalling an old suggestion by @hugomg. This will allow us to annotate the entire AST with line/column information cheaply, if we decide to do so for debugging purposes.

hishamhm (approved):
Looks good! :)

PR #26, opened 2017-10-15
tweaks to the code generator

mascarenhas:
I had to open another branch to edit the README. Also fixed the code generator so it will not export local functions to Lua.


mascarenhas:
By the way, it is not just strings that are missing, global variables too!

hishamhm (approved):

PR #33, opened 2017-10-18
Implementing global variables

mascarenhas:
This pull request is for collecting and discussing the work on implementing global variables for Titan v0.

### DIFF
--- a/examples/artisanal.titan
+++ b/examples/artisanal.titan
@@ -4,6 +4,44 @@ function foo(x: integer, y: integer): integer
     return x + y
 end
 
+function doublea(): integer
+    a = a * 2
+    return a
+end
+
+function doubleb(): integer
+    b = b * 2
+    return b
+end
+
+function setc(t: { integer }): nil
+    c = t
+end
+
+function sumc(): integer
+    local s = 0
+    for i = 1, #c do
+        s = s + c[i]
+    end
+    return s
+end
+
+function sumd(): integer
+    local s = 0
+    for i = 1, #d do
+        s = s + d[i]
+    end
+    return s
+end
+
+local a: integer = foo(2,3)
+
+b: integer = foo(10,5)
+
+local c: {integer} = {1,2,3}
+
+d: {integer} = {3,4,5}
+
 function filltable(N: integer): { integer }
     local xs: { integer } = {}
     for i=1,N do
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -327,12 +327,92 @@ describe("Titan code generator ", function()
         assert.truthy(ok, err)
     end)
 
-    pending("handles coercion to integer", function()
+    it("generates code for integer module-local variables", function()
+        local code = [[
+            local a: integer = 1
+            function geta(): integer
+                return a
+            end
+            function seta(x: integer): nil
+                a = x
+            end
+        ]]
+        local ast, err = parser.parse(code)
+        assert.truthy(ast, err)
+        local ok, err = checker.check(ast, code, "test.titan")
+        assert.truthy(ok, err)
+        local ok, err = generate(ast, "titan_test")
+        assert.truthy(ok, err)
+        local ok, err = call("titan_test", "assert(titan_test.geta() == 1);titan_test.seta(2);assert(titan_test.geta() == 2)")
+        assert.truthy(ok, err)
+    end)
+
+    it("generates code for float module-local variables", function()
+        local code = [[
+            local a: float = 1
+            function geta(): float
+                return a
+            end
+            function seta(x: float): nil
+                a = x
+            end
+        ]]
+        local ast, err = parser.parse(code)
+        assert.truthy(ast, err)
+        local ok, err = checker.check(ast, code, "test.titan")
+        assert.truthy(ok, err)
+        local ok, err = generate(ast, "titan_test")
+        assert.truthy(ok, err)
+        local ok, err = call("titan_test", "assert(titan_test.geta() == 1);titan_test.seta(2);assert(titan_test.geta() == 2)")
+        assert.truthy(ok, err)
+    end)
+
+    it("generates code for boolean module-local variables", function()
+        local code = [[
+            local a: boolean = true
+            function geta(): boolean
+                return a
+            end
+            function seta(x: boolean): nil
+                a = x
+            end
+        ]]
+        local ast, err = parser.parse(code)
+        assert.truthy(ast, err)
+        local ok, err = checker.check(ast, code, "test.titan")
+        assert.truthy(ok, err)
+        local ok, err = generate(ast, "titan_test")
+        assert.truthy(ok, err)
+        local ok, err = call("titan_test", "assert(titan_test.geta() == true);titan_test.seta(false);assert(titan_test.geta() == false)")
+        assert.truthy(ok, err)
+    end)
+
+    it("generates code for array module-local variables", function()
+        local code = [[
+            local a: {integer} = {}
+            function len(): integer 
+                return #a
+            end
+            function seta(x: {integer}): nil
+                a = x
+            end
+        ]]
+        local ast, err = parser.parse(code)
+        assert.truthy(ast, err)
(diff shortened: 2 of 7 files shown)


hishamhm:
By global variables do we mean variables that will be exposed as Lua globals?

I have lots of questions for which I'm not sure we have answers yet -- laying them out here for discussion:

What is the semantics of the module toplevel with regard to variable declarations expected to be?

If we allow `x = 0` at the top level, I'm assuming that means setting a Lua-visible global `x` to `0` at module initialization time; but then what does `local x = 0` mean in the toplevel? With the toplevel not being a chunk, that `local x = 0` is no longer lexically-scoped, which is weird compared to the same statement used within functions (and what would their behavior as upvalues be?) -- nested function declaration has the same issue of course but it's not as evident. OTOH, if we don't allow module-locals, we're promoting usage of globals, which is bad (and slow).

mascarenhas:
By global variables I mean both the ones that are exported (and exposed to Lua as variables in the module) and the ones that are not exported to Lua but visible to the whole Titan module.
Yes, the fact that "local x: integer = 0" at the toplevel makes `x` visible to the whole module is odd, but that is why I am calling these variables (and local functions) "module locals".

Module-local variables are just as fast as function locals, by the way. Variables exported to Lua pay a slight penalty, as they always have to be accessed through a `TValue*` load. All slots used by module-local and exported variables come from the upvalues of the C closure for the __index metamethod of the module table. The module table of a module that exports variables has a metatable (protected against tampering by Lua) that lets Lua code both get and set the values of exported variables.

Right now all of the code for initializing and exporting module variables is in `coder.generate`, this probably needs to be broken up and refactored.

mascarenhas:
I am still leaving this unmerged so we can continue the discussion.

hugomg:
I think that module locals having a different scoping rule than they do in regular Lua wouldnt be that bad as long as the initialization were restricted to side-effect-free expressions (such as literals).

mascarenhas:
@hugomg, we do not have to be that radical, it is safe for the initializers of module variables to call functions from other modules, the problem is calling functions from the module where they are defined, as this lets these functions access unitialized variables. I have pushed a typechecker fix for this.

mascarenhas:
By the way, I am already thinking about how Titan modules `require`-ing other Titan modules will work, that is why I put all initialization for the module under a non-static `<modname>_init` function that has a guard that keeps a module from being initialized twice. The idea is that Titan modules will directly call the initialization function of modules they `require` as part of their initialization.

hishamhm:
> Yes, the fact that "local x: integer = 0" at the toplevel makes x visible to the whole module is odd, but that is why I am calling these variables (and local functions) "module locals".

Yes, I was already seeing them as "module locals" too. This difference in semantics is once again pushing me towards wanting some explicit `module` construct. It also makes me think of a more general principle: perhaps when the semantics are different from Lua, the syntax should reflect this. Given that we envision programmers using both languages back-and-forth a lot, having things that "look the same but act different" is a recipe for tripping up. For example, Lua tables and Titan objects have the syntactic distinction in their type declarations, so we don't have this problem there, but for the toplevel we currently have two very different things looking exactly the same.

hishamhm:
> By global variables I mean both the ones that are exported (and exposed to Lua as variables in the module) and the ones that are not exported to Lua but visible to the whole Titan module.

We need different names for these things, it will be confusing if we call those things "globals" (either in the compiler implementation or in the language reference manual), given the presence of Lua global variables. I suggest restricting the name "global variable" to Lua globals (i.e., fields in Lua's `_G`).

For now I suggest using "module attributes" for exported non-function(*) variables and "module locals" for the non-exported ones.

(* we didn't even open the can of worms yet on to which extent we want to support functions as first-class variables, right? Definitely something for some later time.)

mascarenhas:
I would give up retroactive visibility of module locals to keep using the `local` keyword. But would still make an exception for local module functions, having to do things like `local codestat, codeexp` to have mutually recursive local functions is a Lua wart that always gets on my nerves.

About first-class functions, for the first public release I would probably have them, but only top-level functions, so basically C function pointers. Nested and anonymous functions and consequently true closures can come later.

hishamhm:
> I would give up retroactive visibility of module locals to keep using the local keyword.

I think using `local` for module locals with retroactive visibility would be fine and not-confusing if doing it inside something like a `module` block or declaration, and not a bare-looking toplevel. Unordered declarations are normal when declaring data (e.g. Java classes, Lua table constructors).

Which brings me to another question: what is a module in the Titan type system? We can get away with keeping this question kind of unanswered in v0 but as we move forward with data types it seems to me that modules will unify with structs/objects (with module being a syntactic shorthand for a singleton object with const function static-in-the-Java-sense fields).

> having to do things like local codestat, codeexp to have mutually recursive local functions is a Lua wart that always gets on my nerves.

Same here!

> About first-class functions, for the first public release I would probably have them, but only top-level functions, so basically C function pointers.

I don't even think this is necessary for something called 0.1. But I'd like to have this shortly afterwards.

> Nested and anonymous functions and consequently true closures can come later.

Agreed.

mascarenhas:
> Which brings me to another question: what is a module in the Titan type system? We can get away 
> with keeping this question kind of unanswered in v0 but as we move forward with data types it seems 
> to me that modules will unify with structs/objects (with module being a syntactic shorthand for a
> singleton object with const function static-in-the-Java-sense fields).

In my opinion we should keep it simple like Lua, with 1 source file = 1 module. A module can have functions, variables, classes, type definitions... I do not see the need for a `module` block unless we want to have more than one module in the same source file, but this opens up a can of worms on what does the argument to `require` refers to and the module search path.

Modules would *not* be first class for the time being. With structs that have first-class functions as members or true objects we can fake the kind of "pick one module from this set of modules with the same interface" thing that LuaRocks does.

hishamhm:
> In my opinion we should keep it simple like Lua, with 1 source file = 1 module.

I would prefer that too. I'm conflicted between that and the misleading syntax for the toplevel. And I understand that having a `module` block leads people to want multiple blocks in a file. Would it be too Javaesque if a file could have only one exported module and any number of `local` modules?

> Modules would not be first class for the time being. 

And a `require` call will be magical and only work at the toplevel. Correct?

> With structs that have first-class functions as members or true objects we can fake the kind of "pick one module from this set of modules with the same interface" thing that LuaRocks does.

I'm not sure I follow that part but I'm sure we can get back to this when we get there. :)

mascarenhas:
> I would prefer that too. I'm conflicted between that and the misleading syntax for the toplevel. And I understand that having a module block leads people to want multiple blocks in a file. Would it be too Javaesque if a file could have only one exported module and any number of local modules?

Well, one possibility is to decouple source files and modules. A `foo.titan` source file could declare modules `bar` and `baz` and it would generate `bar.so` and a `baz.so` files at the end. We would need a search path so the compiler could find all the source files. A first pass through the files just collect all the modules that are defined (and check for duplicates), and only modules where the source file is newer than the `.so` file are actually compiled. We could also generate serialized dumps of the definitions (and types) of each module so we don't even need to reparse and typecheck modules that have not changed since the last compilation.

> And a require call will be magical and only work at the toplevel. Correct?

Yes. A semi-crazy idea that I had was to add an `if` statement to the top-level that is strictly compile-time, to enable conditional compilation: the condition must be something that the compiler can evaluate at compile-time (checking module variables where the right-hand side are literal expressions, for example), and the body can only have other top-level definitions.

hishamhm:
> Well, one possibility is to decouple source files and modules. A foo.titan source file could declare modules bar and baz and it would generate bar.so and a baz.so files at the end. We would need a search path so the compiler could find all the source files. A first pass through the files just collect all the modules that are defined (and check for duplicates), and only modules where the source file is newer than the .so file are actually compiled. We could also generate serialized dumps of the definitions (and types) of each module so we don't even need to reparse and typecheck modules that have not changed since the last compilation.

That sounds really interesting. To keep things tidy, I'd suggest that a foo.titan module should only declare modules called either `foo` or `foo.*` (ie, within its namespace). But I like the overall way of thinking of considering the module system with regard to the compilation process.

PR #35, opened 2017-10-21
code generation for strings: literals, concatenation, length, coercion

mascarenhas:
This is the last feature of "Titan v0", after this we can move what is needed to get this from proof of concept to first public release. I started by changing concatenation from a binary expression to n-ary, like Lua does with OP_CONCAT to avoid quadratic behavior.


mascarenhas:
Ok, I wanted to get this done with before heading to SPLASH.

PR #38, opened 2017-10-24
Should we add the benchmarks in this repository?

andremm:
I'm creating this PR to discuss if we should include the benchmarks in this repository. As a start I created a script that runs the `binsearch` benchmark created by @hugomg and then adapted by @mascarenhas to just print out the results for Titan, LuaJIT, and Lua. One nice thing to have one benchmark script is that it is easy to run on different machines. Later on we can improve the script to be smarter and generate the results in a better format instead of just printing them out.

### DIFF
--- a/benchmarks/README.md
+++ b/benchmarks/README.md
@@ -1,0 +1,16 @@
+# Titan benchmarks
+
+This directory contains some bechmarks to compare Titan's execution time
+with Lua and LuaJIT.
+
+# Requirements for running the benchmarks
+
+1. [Lua](http://www.lua.org/) >= 5.3.0
+2. [LuaJIT](http://luajit.org/) >= 2.0.5
+
+# Usage
+
+        $ lua run.lua
+
+The benchmark script `run.lua` uses the Titan compiler `titanc` from the
+parent directory.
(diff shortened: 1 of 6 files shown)


hishamhm:
If we decide to post these in the main repo, we need to add a disclaimer in the benchmarks readme, stating something like "these are microbenchmarks and that they should be taken with a grain of salt".

gligneul:
I prefer to keep things organized and create a separate repo for the benchmarks.

hugomg:
I think having a folder for benchmarks would be organized enough for me. One advantage of having everything under one repo is that the benchmarks can stay synchronized to the current language version

On Oct 24, 2017 2:21 PM, Gabriel de Quadros Ligneul <notifications@github.com> wrote:

I prefer to keep things organized and create a separate repo for the benchmarks.

hishamhm:
We could also use it to track performance regressions eventually (though that can't be done trivially in Travis... unless we find a reliable way to track cycles in there (...running a CPU emulator like QEMU perhaps!?))

mascarenhas:
Shouldn't `perf` work on Travis?

​

hishamhm:
It might, we need to test. It requires CPU accounting options to be enabled in the running kernel, and I don't know if that's exposed in containerized environments. But it's just a matter of testing, really.

PR #40, opened 2017-10-26
First version of the module system

mascarenhas:
Just began working on this, have put the syntax in place for now. Basic idea is that you reference a module with a `local <name> = require <modname>`, where `<name>` is an identifier that will be how the module is referenced inside the current module, and `<modname>` is a string that will tell the Titan compiler where to find the module's source file (using rules similar to Lua search paths).

You reference module members (variables and functions) with `<name>.<name>`, where the first name is the local name that you have given the module on `require`, and the second name is the name of the variable or function. Modules are not first class, and there is no nesting, as you cannot export module references, so `foo.bar.baz` is not a valid reference to a module member. Right now `foo.bar.baz` is just a syntax error, once we have records this will become syntax, but the typechecker will reject it if either `foo` is not a record with a field `bar` that is also a record that has a field `baz`, or `foo` is the local name of an imported module that has a variable `bar` which is a record that has a field `baz`. Easy-peasy. 😄


hishamhm:
As we discussed earlier, I am really concerned about constructs that have identical syntax to Lua but different semantics, especially since we expect programmers to hop back and forth between the two languages.

Our control structures look and behave identically to Lua; our data types do so too when they match Lua types (and are free to behave differently when they don't match Lua types). But our module system works very differently from Lua but looks the same. I really don't want to be trigger-happy about inventing new syntax in Titan (as fun as it sounds at first, it's really hard to get it right so it's always easy to fall back to copying Lua when in doubt) but I really think that different semantics should be visually distinguishable in the source code. It will help programmers make the transition more smoothly ("this is the same, this is different"), rather than feeling like they're walking on thin ice treating it as a weird subset of Lua and guessing what works and what doesn't.

mascarenhas:
> As we discussed earlier, I am really concerned about constructs that have identical syntax to Lua but different semantics, especially since we expect programmers to hop back and forth between the two languages.

What construct are you talking about here? `require` has basically the same behavior in the common case, if you consider that modules (and currently even functions) are not first class. You import a module with `local foo_alias = require "foo"`, and then calls a function `bar` that `foo.titan` defines with `foo_alias.bar(<args>)`. If `foo.titan` defines a module variable `a` you can both read and write to it with `foo_alias.a`. The behavior for neither of those things should be surprising for a Lua programmer. The modules that you import cannot be accessed by anyone that imports you, same as Lua when you `require` a module and assigns it to a local variable.

With the modules themselves I am tooking a few liberties to make them easier to write, and remove some warts that Lua has due to its imperative top-level, but with `require` I was making an effort to copy the semantics that Lua has. We could just as easily have a magical `require "foo"` without a left-hand side that would be equivalent to `local foo = require "foo"`, but I do not think it is worth diverging from Lua for the sake of three less tokens.

hishamhm:
> What construct are you talking about here? require has basically the same behavior in the common case

Yes, I was referring to the `require` keyword. Indeed, it produces the same result in the common case, but it's the edge cases that I'm concerned about. It looks a lot like Lua's require, but it doesn't work like it at all (it's not a function call, etc.) -- people will see `require("foo")` and will assume that you can do with it all that you can do with `require("foo")` in Lua. And then they'll conclude it's a "stripped-down/less-powerful version" of it, when in fact it's not: it's something else entirely, which we'll be able to use to drive the compilation process, eventually manage dependencies, etc.

The module system is a place where we're deviating from Lua from day one, so I think it shouldn't pretend it is a restricted version of Lua's intentionally vague module support.

Not to mention that, since we have Lua integration, users will be able to access Lua's `require` function from Titan, so having two `require`s that behave differently will be really ugly (with Titan's using an assignment that's not really an assignment and a function call that's not really a function call).

Let's just call it `import`, make it an evident keyword, and make a clear distinction between the static `import` and the dynamic `require`.

mascarenhas:
The thing is, if a Lua programmer coming to Titan tries to use one of the edge cases it will not just silently do something unexpected, but fail at compile time with an (hopefully informative) error message. And in time some of these edge cases can be supported by Titan's compile-time `require`, also, given the constraints of it being a compile-time construct. In time it will be important for Titan to have some support for conditional compilation and conditional importing, and some form of compile-time configuration, and I think we can make this work in a way that is familiar to Lua programmers.

> Not to mention that, since we have Lua integration, users will be able to access Lua's require function from Titan, so having two requires that behave differently will be really ugly (with Titan's using an assignment that's not really an assignment and a function call that's not really a function call).

Actually I made `require` a keyword, so the programmer will not be able to call the `require` function that he may have got from Lua `require`.

> Let's just call it import, make it an evident keyword, and make a clear distinction between the static import and the dynamic require.

I worry if this is just added friction from a user coming from Lua. The behavior of `require`, module members and their visibility and restrictions on initialization expressions, and any other compile-time constructs that we may like to support later, in my view all of these follow logically from the fact that Titan's top-level is declarative and most of its effect is compile-time, while Lua's is imperative and its effects are runtime.

Anyway, there is plenty of time for getting feedback from users and see what works and what does not, and we are free to make breaking changes to the language anytime before 1.0, so I think we should focus on getting a usable 0.1 to users ASAP, and them seeing what kind of feedback we get.

hishamhm:
> Actually I made require a keyword, so the programmer will not be able to call the require function that he may have got from Lua require.

Unless they stick it into a table (when we have them. :) ) -- anyway, I think it _would_ be useful to have access to Lua `require` from within Titan functions (when managing the scripting part of a a scriptable application). `require` is our analogue to `dlopen`, while the static import is more like `#include`/`-l`

> The behavior of require, module members and their visibility and restrictions on initialization expressions, and any other compile-time constructs that we may like to support later, in my view all of these follow logically from the fact that Titan's top-level is declarative and most of its effect is compile-time, while Lua's is imperative and its effects are runtime.

Yes, they do follow, I just wanted to make the distinction more evident. My working hypothesis is that things that look-the-same-but-aren't-quite-the-same is more friction:

```
-- How do I require a module?
-- "local foo = require 'foo'" just like Lua.
-- Nice! Thanks!

...a while later...

-- Hey, "local ok, foo = pcall(require, 'foo')" doesn't work!
-- Well, we sort of lied, you weren't using the require function you thought you were using...
-- (╯°□°)╯︵ ┻━┻
```

versus

```
-- How do I require a module? I tried require and it doesn't work
-- Ah, in Titan you use "import 'foo'": you're importing statically
-- Oh well, thanks! ¯\_(ツ)_/¯
```

hugomg:
I think having `local foo = require "foo"` just work would also avoid a lot of friction though... I would also expect programmers to understand that Titan is less dynamic than Lua and that not every Lua idiom can be ported intact.

I would generally be OK with Titan having more a more restrictive version of the Lua semantics, in that valid titan programs produce the same result as Lua would, but some valid Lua programs might result in a compile or runtime error if you try to run them as Titan. (This is the Gradual Guarantee that the gradual typing folk are also fond of). I'm more concerned about the exceptions to this rule. With require in particular, my concern would be ensuring that Titan's require will find the same modules that Lua's require would, so that converting a Lua program to Titan by adding some type annotations doesn't break the imports. (Perhaps we might need  a Titan-friendly package.path or package loader to do this?)

mascarenhas:
> With require in particular, my concern would be ensuring that Titan's require will find the same modules that Lua's require would, so that converting a Lua program to Titan by adding some type annotations doesn't break the imports. (Perhaps we might need a Titan-friendly package.path or package loader to do this?)

@hugomg is very worried about this, but I repeatedly tell him that I do not see what the issue is. 😄 If you `require` (or `import`, whatever we decide to go with) a Titan module you go through the (compile-time) search path where `titanc` will find your `.titan` file. This is a mostly compile-time thing (the runtime system might do some initialization of your module at runtime, but actually loading the `.so` is fully up to the dynamic linker of your OS). If you `require` a Lua module there is nothing `titanc` does at compile-time, as it is a black box, at runtime this is resolved using whatever you have in the `package.loaders` of the Lua interpreter that you are using. Even if we go with Typed Lua-style declaration files for Lua modules, to have better autocompletion, it still does not matter what the actual module is at runtime, because `titanc` cannot trust the types you put in the declaration file anyway. I do not see at any point in time `titanc` actually having to go and load a Lua module *at compile-time*, which could lead to the problem that @hugomg has in mind (the compiler sees a module, at runtime it is another).

Notice that if a Titan module uses another Titan module this **does not** go through Lua's module loader, the `.so` of the first module actually has a direct reference to the `.so` of the other module, and this is resolved by you operating system. The user can still get in DLL hell if he is not careful, though, but this has nothing to do with Lua.

hishamhm:
> I'm more concerned about the exceptions to this rule.

I am strongly opposed to this ever being a rule in the first place: Titan was never meant to be a subset of Lua. Given the universe of Lua programs, the vast majority of them won't ever be valid Titan programs. I remain fully skeptical of the "gradual" way of thinking. Approaching its design from this perspective of "Titan as valid Lua programs" will only result in situations like the current `require` with an assignment-that's-not-an-assignment, function-call-that's-not-a-function-call, etc. I'm afraid we're shoehorning one language's syntax into another language's semantics and producing confusion in the process. Look at @mascarenhas's lenghty explanation to @hugomg in the comment above about the different behaviors of Titan's `require`. If these differences are blurry even to the dev team, imagine to users later. My case could not have been more clearly stated.

mascarenhas:
> Look at @mascarenhas's lenghty explanation to @hugomg in the comment above about the different behaviors of Titan's require. If these differences are blurry even to the dev team, imagine to users later. My case could not have been more clearly stated.

I am not fully convinced, but point taken. 😃 We can change it to `import` for now, and then revisit this discussion after we get feedback from users.

andremm:
>> Look at @mascarenhas's lenghty explanation to @hugomg in the comment above about the different behaviors of Titan's require. If these differences are blurry even to the dev team, imagine to users later. My case could not have been more clearly stated.

>I am not fully convinced, but point taken. 😃 We can change it to import for now, and then revisit this discussion after we get feedback from users.

I'm glad we are giving a chance to `import`, as I totally agree with @hishamhm on this. To my mind, 
we should try avoiding this feeling that Titan looks like a gradual typing Lua-like dialect, as this might bring some confusion among users, and creating our own way for handling modules is the first thing we could do to try avoiding misunderstandings.

hishamhm:
@mascarenhas Thanks, Fabio. I just pushed a commit to do the rename. I tried to keep the changes as small as possible, using the `local bar = import "foo.bar"` syntax. (I just removed suport for the parenthesized variant `import("foo.bar")` to make it clear it's not a function call.)

hishamhm:
Personally I'd like it if `import "foo.bar"` was sugar for `local bar = import "foo.bar"`, to discourage users from renaming modules needlessly (something that happens in Lua all too often), but I thought that would be pushing it :) (keep in mind however that function declarations do have a sugared form in Lua, though, so that wouldn't be a first in the toplevel).

mascarenhas:
I was already thinking about that feature, to be honest, with exactly the
behavior that you are suggesting. :-)

Em 1 de nov de 2017 6:30 PM, "Hisham Muhammad" <notifications@github.com>
escreveu:

hishamhm:
I took the liberty of making a cleanup since this PR contained a number of unrelated commits. I pushed the unrelated ones to `master` and the other ones into PR #44 so we can continue from there!

PR #41, opened 2017-10-26
Make sure the generated code indentation is consistent

hishamhm:
Reindent the generated code to make sure the indentation of
the output is correct.

### DIFF
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -208,7 +208,7 @@ local function codeif(ctx, node, idx)
     if idx == #node.thens then -- last condition
         cstats, cexp = codeexp(ctx, node.thens[idx].condition, true)
         cthn = codestat(ctx, node.thens[idx].block)
-        cels = node.elsestat and "else " .. codestat(ctx, node.elsestat) or ""
+        cels = node.elsestat and "else " .. codestat(ctx, node.elsestat):match("^[ \t]*(.*)") or ""
     else
         cstats, cexp = codeexp(ctx, node.thens[idx].condition, true)
         cthn = codestat(ctx, node.thens[idx].block)
@@ -219,7 +219,8 @@ local function codeif(ctx, node, idx)
             %s
             if(%s) {
                 %s
-            } %s
+            }
+			%s
         }
     ]], cstats, cexp, cthn, cels)
 end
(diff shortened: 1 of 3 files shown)


titan-compiler/pretty.lua
```diff
@@ -6,7 +6,7 @@ function pretty.reindent_c(input)
     local blank = false
     for line in input:gmatch("([^\n]*)") do
         local do_print = true
-        line = line:match("^[ \t]*(.*)[ \t]*$")
+        line = line:match("^[ \t]*(.-)[ \t]*$")
```
hishamhm: Thanks! :)

hugomg:
I like how this code is much simpler than what I would have done :)

It doesn't do the right thing if we write things like `if(blah) {}` though (lowers indent level but should leave it unchanged). Perhaps we should add a comment clarifying this?

hishamhm:
> It doesn't do the right thing if we write things like if(blah) {}

Are you sure? I only match `{$` to indent and `^}` to unindent. The example you gave doesn't match either.

hugomg:
Oops, I guess you are right.

This means that the only corner case should be curly brackets inside comments but that should never come up for us :)

andremm (approved):

titan-compiler/coder.lua
```diff
@@ -219,7 +219,8 @@ local function codeif(ctx, node, idx)
             %s
             if(%s) {
                 %s
-            } %s
+            }
+			%s
```
hishamhm, line 223: @andremm was the `reindent_c` function misbehaving in this case? I thought it handled `} else {` correctly since it checks for both `^}` and `{$` in the same pass. I'd rather fix the reindenter than to have `else` in a separate line. :)

andremm: > @andremm was the reindent_c function misbehaving in this case? I thought it handled }  else { correctly since it checks for both ^} and {$ in the same pass. I'd rather fix the reindenter than to have else in a separate line. :)

sorry @hishamhm, I thought that applying this patch would be easier than having special cases in the reindenter. So, I applied it because leaving `} %s` generates lines containing the pattern `}\t\t{`, and leaving `"else " .. codestat(ctx, node.elsestat)` without left trim generates lines containing the pattern `else \t{`.
Should we roll back and fix that in the reindenter?

andremm: Anyway, this has been rolled back after fixing the generation of if statements on PR #40.
Now we do not have `else` in a separate line anymore. :)

PR #42, opened 2017-11-01
`for-in` construct

hishamhm:
Just pushed to the server some code I had lying around when I started playing with adding `for-in`. So far it has just the parser update and beginnings of type checking.

Evidently this needs to wait until we have other features in place for the complete functionality of `for-in` to work (first class functions being the most obvious one) unless we want to support a hacky version first with hardcoded support for the standard library iterators only (which I see as being special-cased in the code-generator anyway, so it's not like that would be throwaway code).


mascarenhas:
Having the possibility for top-level functions being first-class functions, which is enough for Lua's stateless iterators, is not hard, as they are literally function pointers. But to really be convenient we also need to have parametric polymorphism, otherwise you need to have a different ipairs for each type of array element.

PR #43, opened 2017-11-01
Detect attempts to call non-functions

hishamhm:
Allow calling only functions for now. This will change when
we get other callable objects.


mascarenhas (approved):
Looks good, thanks! I thought I had already done that on my first pass through the type checker, but looks like it got overlooked.

PR #44, opened 2017-11-01
Ongoing work on the module system

hishamhm:
This is a continuation of PR #40 (see there for the initial round of discussion!)

I cherry-picked each commit from that branch and committed the ones unrelated to modules directly into `master`, and all module-related ones in this new clean branch. This way we get a clean branch for merging (since that PR had merges back from master).


mascarenhas:
There is a very nasty conflict on `checker_spec.lua`, can someone look into that and fix it before doing more work on this branch so the conflict does not get worse?

hishamhm:
@mascarenhas I'm on it. Please see PR #50 -- I'll need to rebase this, but it will be a smaller, stabler diff.

hishamhm:
Fixed the merge conflicts, this branch is good to work on again. :)

I didn't realize that coder_spec.lua had the same pattern adding an argument to all tests, or else I would have refactored that too... :\

mascarenhas:
Can we merge this ASAP? I just spent the last hour or so fixing a VERY NASTY conflict with @gligneul's work on parsing records, and this is bound to happen again when he reopens the branch to work on typechecking and code generation for records concurrently with what I still need to do here.

By the way, I merged master into this branch with "rebase", but handling conflicts seems to be much easier if I just do "merge". Can anyone with more experience with this kind of git workflow advise?

mascarenhas:
By the way, instead of long-lived feature branches such as what this one had become I think the better approach is to aggresively merge back into master as long as the tests are passing, even if there is still work to do on that feature (this is how the work on records has been handled so far). This way we can all stay in sync.

mascarenhas (approved):
I am taking the liberty to review this myself. I think rebase left the diff too dirty to see what has been actually added by this branch...

PR #46, opened 2017-11-03
Adding more tests

andremm:
This PR is intended to be used for adding more tests. Also, I would like to discuss how should be our standard way to add more than just one test for the same thing being tested. For instance, for detecting non-nil functions we had just one test:
```lua
it("detects nil returns on non-nil functions", function()
        local code = [[
            function fn(): integer
            end
        ]]
        local ast, err = parser.parse(code)
        local ok, err = checker.check(ast, code, "test.titan")
        assert.falsy(ok)
        assert.match("function can return nil", err)
end)
```
but then I added more tests by simply duplicating code:
```lua
    it("detects nil returns on non-nil functions", function()
        local code = [[
            function fn(): integer
            end
        ]]
        local ast, err = parser.parse(code)
        local ok, err = checker.check(ast, code, "test.titan")
        assert.falsy(ok)
        assert.match("function can return nil", err)
        code = [[
            function getval(a:integer): integer
                if a == 1 then
                    return 10
                elseif a == 2 then
                else
                    return 30
                end
            end
        ]]
        ast, err = parser.parse(code)
        ok, err = checker.check(ast, code, "test.titan")
        assert.truthy(err)
        assert.match("function can return nil", err)
        code = [[
            function getval(a:integer): integer
                if a == 1 then
                    return 10
                elseif a == 2 then
                    return 20
                else
                    if a < 5 then
                        if a == 3 then
                            return 30
                        end
                    else
                        return 50
                    end
                end
            end
        ]]
        ast, err = parser.parse(code)
        ok, err = checker.check(ast, code, "test.titan")
        assert.truthy(err)
        assert.match("function can return nil", err)
    end)
```
In this PR I moved them to use `for` along with `ipairs` instead:
```lua
    it("detects nil returns on non-nil functions", function()
        local code = {[[
            function fn(): integer
            end
        ]],
        [[
            function getval(a:integer): integer
                if a == 1 then
                    return 10
                elseif a == 2 then
                else
                    return 30
                end
            end
        ]],
        [[
            function getval(a:integer): integer
                if a == 1 then
                    return 10
                elseif a == 2 then
                    return 20
                else
                    if a < 5 then
                        if a == 3 then
                            return 30
                        end
                    else
                        return 50
                    end
                end
            end
        ]],
        }
        for _, c in ipairs(code) do
            local ast, err = parser.parse(c)
            local ok, err = checker.check(ast, c, "test.titan")
            assert.falsy(ok)
            assert.match("function can return nil", err)
        end
    end)
```
Or should we create one single test for each little program like below?
```lua
it("detects nil returns on non-nil functions", function()
        local code = [[
            function fn(): integer
            end
        ]]
        local ast, err = parser.parse(code)
        local ok, err = checker.check(ast, code, "test.titan")
        assert.falsy(ok)
        assert.match("function can return nil", err)
end)
it("detects nil returns on non-nil functions", function()
        local code = [[
            function getval(a:integer): integer
                if a == 1 then
                    return 10
                elseif a == 2 then
                else
                    return 30
                end
            end
        ]]
        local ast, err = parser.parse(code)
        local ok, err = checker.check(ast, code, "test.titan")
        assert.falsy(ok)
        assert.match("function can return nil", err)
end)
it("detects nil returns on non-nil functions", function()
        local code = [[
            function getval(a:integer): integer
                if a == 1 then
                    return 10
                elseif a == 2 then
                    return 20
                else
                    if a < 5 then
                        if a == 3 then
                            return 30
                        end
                    else
                        return 50
                    end
                end
            end
        ]]
        local ast, err = parser.parse(code)
        local ok, err = checker.check(ast, code, "test.titan")
        assert.falsy(ok)
        assert.match("function can return nil", err)
end)
```
I like the `for` way, by the way.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -43,7 +43,7 @@ describe("Titan type checker", function()
         assert.same("Exp_ToFloat", ast[1].block.stats[2].exp._tag)
     end)
 
-    it("catches duplicate declartions", function()
+    it("catches duplicate function declarations", function()
         local code = [[
             function fn(): integer
                 return 1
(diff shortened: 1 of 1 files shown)


hishamhm:
We can use `for`, but each test should have its own `it(name, fn)` call, so it becomes accounted separately in Busted. So, one way would be to have
```lua
local tests = {
    { name = "foo", code = [[ .. ]] },
    { name = "bla", code = [[ .. ]] },
}
for _, test in ipairs(tests) do
    it(test.name, function()
        local ast, err = parser.parse(test.code)
        local ok, err = checker.check(ast, c, "test.titan")
        assert.falsy(ok)
        assert.match("function can return nil", err)
    end
end
```

Having separate names is also nice because it makes it clearer what's the case you are testing (`empty elseif case`, `nested statements`, etc).

hugomg:
In the test files I wrote I solved this problem by [putting the repeated code inside a function](https://github.com/titan-lang/titan-v0/blob/cdc2cc9495ba17f97928955be2d642b42ac815cb/spec/lexer_spec.lua#L59). But I think a for loop would be fine too.

BTW, what would you consider "each test case" to be, Hisham? For example, in lexer_spec I made [many it blocks](https://github.com/titan-lang/titan-v0/blob/cdc2cc9495ba17f97928955be2d642b42ac815cb/spec/lexer_spec.lua#L71) that contain multiple test cases inside them, because I felt that the line number in the stack trace was enough to tell them apart from each other.

hishamhm:
Besides better documenting each case that is being tested, practical advantages of isolating each test in its own test function are:
* all of them run -- a test with multiple asserts will stop at the first one. So many separate tests gives you a better view of what's broken, when you're testing multiple test cases 
* it helps to ensure that tests have no interdependencies/side-effects (e.g. busted can run them in random order)

andremm:
I like this approach @hishamhm!

hishamhm (approved):
I retriggered the Travis tests after #47 was merged and now the `pr` test passed.

If you want to refactor the tests to generate separate `if()` entries, feel free, but as @hugomg said we could in addition create some utility functions to avoid repetition, so maybe a separate commit with the refactoring would be best. So, in spirit of incremental improvements, I'm giving this one a +1 as is :)

PR #47, opened 2017-11-03
marking a variable as assigned only if it was declared

andremm:
The following test case makes the Titan compiler crash:
```lua
function fn(): nil
    local x: integer = 1
    y = 2
end
```
This PR fixes this problem and it is necessary for two test cases that are failing on PR #46. One of these test cases is the above example.


mascarenhas (approved):

hishamhm:
sweet :)

PR #48, opened 2017-11-03
tentative roadmap

mascarenhas:


hishamhm:
Looks good! I just wonder about the version numbers, since, if we assume semantic versioning, having them all `1.x` means that they keep full backward compatibility. Rust and Go took their time to reach `1.0` and start giving compatibility promises — I wonder if we shouldn't do the same.

mascarenhas:
Yep, the version numbers themselves are a big part of the "tentative" part of the roadmap. They can be a bunch of v0.x, v0.(x+1), v0.(x+2) etc., or v1.0, v1.1, v2.0, v2.1, etc. languages usually take the first route and move slowly once they go 1.0, there are even languages like Elm that are already pretty popular and used in production and still at 0.x (Elm is at 0.18). The counterexample is probably Swift, which is already at Swift 4 after just three years from Swift 1.0.

hishamhm:
I think we can be "conservative" at version numbers and use `0.x` as we build up the language.

> The counterexample is probably Swift, which is already at Swift 4 after just three years from Swift 1.0.

One has to take into account that Swift was an internal project at Apple for 4 years before the announcement, so they had their long `0.x` road as well. :)

mascarenhas:
> I think we can be "conservative" at version numbers and use 0.x as we build up the language.

Ok, but we do not have to go with a public release as v0.1, as a lot of the work that we have already done (and have to do to get this first release ready) counts as point releases, in my view. I pushed some tentative tags on what I thought were important milestones so far, and updated the roadmap accordingly.

hishamhm:
There is a _lot_ of work to put into each release of these milestones listed in the roadmap. Each of these bullets is a lot more functionality than the delta added in the v0.1-v0.5 you tagged (I'm not even sure these should have been marked as versions). Having feature targets for numbered releases adds a lot of anxiety. Let's bump 0.x numbers freely as we add user-visible features. 

I suggest we either:

* remove the version numbers from the README altogether
* remove version numbers from the bullets and write it so that the language at the end of that roadmap will be Titan 1.0.
* make each bullet there a major x.0 version.

hishamhm:
A for the current 0.x tags, let's leave them as is, but let's add a CHANGELOG.md file listing what's new/changed in each of these versions.

mascarenhas:
> There is a lot of work to put into each release of these milestones listed in the roadmap. Each of these bullets is a lot more functionality than the delta added in the v0.1-v0.5 you tagged (I'm not even sure these should have been marked as versions). Having feature targets for numbered releases adds a lot of anxiety. Let's bump 0.x numbers freely as we add user-visible features.

The tags are just a first suggestion, they can be changed freely. Could just as well be just 0.1 (the state of the repository at the time of the Lua workshop) and 0.2 (strings, basically the scope of my original "v0" email), modules can be 0.3, with the first public release in the roadmap 0.5. I just did not want to tag that first release as 0.1, as we want it to be a basic but quite usable language.

> remove version numbers from the bullets and write it so that the language at the end of that roadmap will be Titan 1.0

In all, this is probably the best suggestion. We can tag the first public release with an arbitrary v0.X number, and keep bumping it freely on the way to 1.0 when adding new features.

hishamhm:
> In all, this is probably the best suggestion. We can tag the first public release with an arbitrary v0.X number, and keep bumping it freely on the way to 1.0 when adding new features.

+1!

hishamhm:
@mascarenhas Please check my review of the roadmap — I made it focused on the feature order towards 1.0, not version release milestones.

hishamhm (approved):

andremm:
It would be create to have a change log where we can check which features were included or removed from one version to another.

hishamhm:
@andremm yes, that would be a welcome addition too. I'd just like to check with @mascarenhas if the roadmap is good to merge as is before hitting the green button. :)

PR #49, opened 2017-11-09
Start of bare-bones reference manual

mascarenhas:
What it says on the tin. :-)

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -1,0 +1,14 @@
+# Titan Language Reference
+
+## Arrays
+
+Array types in Titan have the form `{ t }`, where `t` is any Titan type (including other array types, so `{ { integer } }`
+is the type for an array of arrays of integers, for example. 
+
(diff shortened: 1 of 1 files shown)


hishamhm (approved):

hishamhm:
Yay! I actually learned something new reading it! :)

hugomg:
Is there a particular reason why arrays default to `{integer}`?

mascarenhas:
> Is there a particular reason why arrays default to {integer}?

Just following C as having `integer` be the default type. When we have the `lua` dynamic type it will be a better default.

It would be easy enough to have the type be `{ <unbound type variable> }` and then unify with something else when more context appeared, but things would get messy pretty fast with subtyping, and there would always be the temptation to go full H-M and dig ourselves a hole. 😄

hishamhm:
+1 to the approach of documenting the language as the implementation currently is, and not what the language should become.

PR #53, opened 2017-11-10
allowing relational expressions for comparing strings

andremm:
The following example was not type checking:
```lua
function fn(a: string, b: string):boolean
    return a < b
end
```
This PR fixes the type checker to allow comparing strings with relational operators and adds some type checker tests to the following binary operations: `==`, `~=`, `<`, `>`, `<=`, and `>=`.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -531,5 +531,256 @@ describe("Titan type checker", function()
         local ok, err = run_checker(code)
         assert.truthy(ok)
     end)
+
+    for _, op in ipairs({"==", "~="}) do
+        it("can compare arrays of same type using " .. op, function()
+            local code = [[
+                function fn(a1: {integer}, a2: {integer}): boolean
+                    return a1 ]] .. op .. [[ a2
+                end
+            ]]
+            local ast, err = parser.parse(code)
+            local ok, err = checker.check(ast, code, "test.titan")
+            assert.truthy(ok)
+        end)
+    end
+
+    for _, op in ipairs({"==", "~="}) do
+        it("can compare booleans using " .. op, function()
+            local code = [[
+                function fn(b1: string, b2: string): boolean
+                    return b1 ]] .. op .. [[ b2
+                end
+            ]]
+            local ast, err = parser.parse(code)
+            local ok, err = checker.check(ast, code, "test.titan")
+            assert.truthy(ok)
+        end)
+    end
+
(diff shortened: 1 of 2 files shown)


spec/checker_spec.lua
```diff
@@ -531,5 +531,256 @@ describe("Titan type checker", function()
...
+
+    for _, op in ipairs({"==", "~="}) do
+        it("can compare arrays of same type using " .. op, function()
+            local code = [[
+                function fn(a1: {integer}, a2: {integer}): boolean
+                    return a1 ]] .. op .. [[ a2
+                end
+            ]]
+            local ast, err = parser.parse(code)
+            local ok, err = checker.check(ast, code, "test.titan")
```
hishamhm: Please use the `run_checker` function instead of calling `parser.parse` and `checker.check` — in PR #44 there is a change to `checker.check`, so using the helper function will abstract away that change and avoid conflicts

hishamhm (changes requested):

spec/checker_spec.lua
andremm: Sorry, I just copied and pasted from an old file I had being working previously.

hishamhm (approved):

hishamhm:
@andremm No problem!

Are you going to work on the code generation for these comparisons as well?

I think we need to come up with a strategy for testing type checking and code generation, because most (all?) of the test cases that type-check need to go to the code-generation tests as well... we're ending up with a lot of duplication.

mascarenhas:
> I think we need to come up with a strategy for testing type checking and code generation, because most (all?) of the test cases that type-check need to go to the code-generation tests as well... we're ending up with a lot of duplication.

I am against merging the tests, as they *are* testing different things... some duplication in the specific Titan code that is tested by the tests does not bother me, as they will only change if the syntax of the Titan fragment that they use changes, which should be rare.

One thing that I want to do now that we have a full lex->parse->typecheck->codegen pipeline is to drive it all with a `compile` function that takes a source file name and a loader function from file names to source strings (so we can compile collections of Titan modules in our tests without having to set those things up in the disk), and returns a table with the results: errors, warnings, asts for all the compiled modules, and compiled code for the modules that compiled without errors.

hishamhm:
Just to make it clear, I did not mean merging the tests, but making sure that the code gen tests cover the same ground of the language spec that the type checking tests do.

I did some experiments in the past in another never-published project of mine where I did some instrumentation to ensure test coverage based on node types in addition to lines. That worked well, I may try to do something similar for Titan.

PR #55, opened 2017-11-10
Barebones template function for code generation

hugomg:
We replaced all uses of string.format for generating C code with a little function that used named placehoders instead of positional placeholders.

There are also some very minor code cleanups (whitespace fixes, C literal creation, and C identifier creation)

----

Bikesheding moment: I don't like the "output" name for the function because it's ambiguous about being a noun or verb and because it isn't specific to what the function does. But I couldn't think of something better :/

### DIFF
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -4,6 +4,15 @@ local coder = {}
 
 local codeexp, codestat
 
+local function output(code, vars)
+    local substitutions = setmetatable(vars, {
+        __index = function(_, k)
+            error("Internal compiler error: missing template variable " .. k)
+        end
+    })
+    return (string.gsub(code, "$([%u%d_]+)", substitutions))
+end
+
 local function quotestr(s)
     s = s:gsub("\\", "\\\\")
         :gsub("\a", "\\a")
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -516,7 +516,7 @@ local function codeassignment(ctx, node)
                     _slot = (TValue *)luaH_getint(_t, _k);
                     TValue _vk; setivalue(&_vk, _k);
                     if (_slot == luaO_nilobject)    /* no previous entry? */
-                            _slot = luaH_newkey(L, _t, &_vk);    /* create one */
+                        _slot = luaH_newkey(L, _t, &_vk);   /* create one */
                 }
                 $CSET
             }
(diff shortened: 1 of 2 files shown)


titan-compiler/coder.lua
```diff
@@ -238,81 +338,101 @@ local function codefor(ctx, node)
...
+                cstep = output(tmpl, subs)
+                ccmp = output("$CVAR <= _forlimit", subs)
             else
+                subs.NEGILIT = c_integer_literal(-ilit)
                 if types.equals(node.decl._type, types.Integer) then
-                    cstep = node.decl._cvar .. " = l_castU2S(l_castS2U(" .. node.decl._cvar .. ") - " .. tostring(-ilit) .. ")"
+                    cstep = output("$CVAR = l_castU2S(l_castS2U($CVAR) - $NEGILIT)", subs)
                 else
-                    cstep = node.decl._cvar .. " -= " .. tostring(ilit)
+                    cstep = output("$CVAR -= $ILIT", subs)
```
hishamhm: Re-reading the code I just spotted a bug (that I wrote myself)... ILIT is not defined in subs at this point. The fact that this wasn't caught by Busted means we don't have a test covering this line!

titan-compiler/coder.lua
```diff
@@ -550,46 +722,75 @@ local function codetable(ctx, node, target)
     local stats = {}
     local cinit, ctmp, tmpname, tmpslot
     if target then
+        -- TODO: double check this code, it wan't covered by tests
+        -- and wan't passing anything to the second $TMPNAME placeholder
```
hishamhm, line 728: Good catch! The new template function will avoid these things in the future :)

titan-compiler/coder.lua
```diff
@@ -625,16 +826,25 @@ local function codebinaryop(ctx, node, iscondition)
...
-                %s = %s;
-                if(%s) {
-                  %s  
-                  %s = %s;
+            local code = output([[
+                $LSTATS
+                $CTMP
+                $TMPNAME = $LCODE;
+                if($TMPNAME) {
+                  $RSTATS
```
hishamhm, line 836: (nitpick: the C code templates above were intented with 4 spaces)

hishamhm:
> Bikesheding moment: I don't like the "output" name for the function because it's ambiguous about being a noun or verb and because it isn't specific to what the function does. But I couldn't think of something better :/

@hugomg A valid concern, though it didn't seem ambiguous to me because I try to make all function names to be verbs. I was out of ideas for alternatives myself, so I outsourced the brainstorming: 😄 

https://twitter.com/hisham_hm/status/928977889265143809

titan-compiler/coder.lua
hishamhm: Pushed a fix for this, and tests

hishamhm (approved):

PR #56, opened 2017-11-10
add GC checks to loops and function preambles

mascarenhas:
Just discovered that Lua does not do a GC step on the internal object allocation primitives, but on the *opcodes* that allocate (as well as C API functions that call those primitives), so Titan was not actually doing any GC while running. Plugged this hole by adding calls to `luaC_checkGC` to loop bodies and function preambles, while we think of a better policy, and how to be smarter about avoiding adding those calls to loops that do not do any allocation.

### DIFF
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -264,16 +264,19 @@ local function codewhile(ctx, node)
         tmpl = [[
             while($CEXP) {
                 $CBLK
+                $CHECKGC
             }
         ]]
     else
         tmpl = [[
             while(1) {
+                luaC_checkGC(L);
                 $CSTATS
                 if(!($CEXP)) {
                     break;
                 }
                 $CBLK
+                $CHECKGC
             }
         ]]
     end
@@ -281,6 +284,7 @@ local function codewhile(ctx, node)
         CSTATS = cstats,
         CEXP = cexp,
         CBLK = cblk,
+        CHECKGC = ctx.nslots > 0 and "luaC_checkGC(L);" or ""
     })
 end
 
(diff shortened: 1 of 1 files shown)


mascarenhas:
Also, any suggestions on how we could test that?

hugomg:
It would be non trivial work but perhaps we could take a similar approach to what Lua does. There are compile-time flags that add extra asserts to the GC-related code and some test cases that are built to stress the GC.

titan-compiler/coder.lua
```diff
@@ -264,23 +264,27 @@ local function codewhile(ctx, node)
         tmpl = [[
             while($CEXP) {
                 $CBLK
+                $CHECKGC
             }
         ]]
     else
         tmpl = [[
             while(1) {
+                luaC_checkGC(L);
```
hugomg: Why do we need a luaC_checkGC() here and why isn't it also using $CHECKGC?

hishamhm:
Can't we emit the GC call on our operations that allocate, instead of the loops? I remember seeing stuff like `isgc` in various places.

mascarenhas:
We could, but I think it might be overkill. We have to do some testing to
find out how to best tune the GC to Titan code. This also enables some
optimizations around whether or not to push things on the Lua stack.

Em 10 de nov de 2017 7:14 PM, "Hisham Muhammad" <notifications@github.com>
escreveu:

titan-compiler/coder.lua
hishamhm: @mascarenhas this is just a slip, right? I think it's the only pending issue before merging this PR.

mascarenhas: Yes, this was an oversight, this line should be removed, as there is already a $CHECKGC at the end of the loop.

mascarenhas:
Ok, removed the extraneous GC check at the start of some while loops. Right now it is not really 100% safe to exploit the fact that we are not calling the GC on every allocation to try to avoid pushing things on the Lua stack when not really needed, because Lua can always do an emergency GC when allocating memory, but the emergency GC can be turned off as a configuration option.

hishamhm:
For one moment I forgot we're bypassing the C API and thought "but how can we allocate _without_ going through the stack?" :)

hishamhm (approved):

PR #58, opened 2017-11-10
Add missing modules to rockspec

hugomg:
Fixed an issue where running titanc outside the titan git repository
would result in a "module not found" error for titan-compiler.types

Question for the rockspec experts: is editing the existing rockspec file the way to go or should I have created a new one?


hishamhm:
In the repo we can just edit. We only increase tge revision number for ones that are published in luarocks.org.

hishamhm (approved):

Issue #59, opened 2017-11-14
Integer overflow in for loops


hishamhm:
bug reported in the Lua list:
http://lua-users.org/lists/lua-l/2017-11/msg00053.html
Lua error case:
```lua
-- print biggest 8 integers on Lua 5.3.4
for i = 0x7FFFFFFFFFFFFFF8,0x7FFFFFFFFFFFFFFF do
  print(i)
end
```

This goes in an endless loop.

Reproduced in Titan:
```lua
-- bignumbers.titan
function bignumbers(nth: integer):integer
    local c = 0
    for i = 0x7FFFFFFFFFFFFFF8,0x7FFFFFFFFFFFFFFF do
        c = c + 1
        if c == nth then
            return i
        end
    end
    return 0
end
```
Test from Lua:
```lua
Lua 5.3.3  Copyright (C) 1994-2016 Lua.org, PUC-Rio
> big = require("bignumbers")
> big.bignumbers
function: 0x7f5cbc0cb040
> print(big.bignumbers(1))
9223372036854775800
> for i = 1, 10 do print(big.bignumbers(i)) end
9223372036854775800
9223372036854775801
9223372036854775802
9223372036854775803
9223372036854775804
9223372036854775805
9223372036854775806
9223372036854775807
-9223372036854775808
-9223372036854775807
> 
```

The last two numbers should be 0 and not negative numbers.

PR #60, opened 2017-11-14
Handle integer overflow in for loops

hishamhm:
Test case for issue #59.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -18,7 +18,7 @@ matrix:
           packages:
             - g++-5
       env:
-        - USE_GCC5="CC=gcc-5 && CXX=g++-5"
+        - CC="gcc-5"
         - INSTALL_LUA="lua=5.3"
         - LUA="lua"
 #    - os: linux
@@ -56,7 +56,7 @@ matrix:
 #        - LUA="luajit"
 
 before_install:
-  - eval "${USE_GCC5}"
+  - export CC=$CC
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
   - pip install hererocks
   - hererocks lua_install -rlatest --$INSTALL_LUA
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,14 @@ matrix:
 #        - INSTALL_LUA="lua=5.2"
 #        - LUA="lua"
     - os: linux
+      addons:
+        apt:
+          sources:
+            - ubuntu-toolchain-r-test
+          packages:
+            - g++-5
       env:
+        - USE_GCC5="CC=gcc-5 && CXX=g++-5"
         - INSTALL_LUA="lua=5.3"
         - LUA="lua"
 #    - os: linux
@@ -49,6 +56,7 @@ matrix:
 #        - LUA="luajit"
 
 before_install:
+  - eval "${USE_GCC5}"
   - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update; fi
   - pip install hererocks
   - hererocks lua_install -rlatest --$INSTALL_LUA
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -26,6 +26,16 @@ local function call(modname, code)
     return os.execute(cmd)
 end
 
+local function compile(titan_code, lua_testcase)
+    local ast = parser.parse(titan_code)
+    local ok, err = checker.check(ast, titan_code, "test.titan")
+    assert.truthy(ok, err)
+    ok, err = generate(ast, "titan_test")
+    assert.truthy(ok, err)
+    ok, err = call("titan_test", lua_testcase)
+    return ok, err, ast
+end
+
 describe("Titan code generator", function()
     it("deletes array element", function()
         local code = [[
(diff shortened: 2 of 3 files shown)


hishamhm:
(Note: the above test-case is 64-bit specific, I think it's the first one. We never discussed what's the plan for Titan in 32-bit machines, but we're using `lua_Number` and `lua_Integer` for now. I think we should just keep using that to keep assumptions at a minimum but not dive too much in the Lua number support matrix at this point, since we already have a lot on our plate — we can rework tests etc to 32-bit when we see the actual need (i.e. users needing those platforms).)

mascarenhas:
Please squash the for commits into a single one when merging! Three of them were just to get everything passing on Travis.

hishamhm:
This solution is nice, but its consequences are far-reaching:

* the generated code is no longer 100% portable
* even for GCC it requires a specific version

Also, the change in semantics is not to be taken lightly, but opening up this issue I was aware of that. If we adopt this as an optimization, is it still possble to do a portable fallback?

In more concrete terms, would changing the semantics of
```lua
for i = x, y, step do
   k(i)
end
```
from
```lua
do -- simplified for positive step only
   local i = x
   while i < y do
      k(i)
      i = i + step
   end
end
```
to
```lua
do -- simplified for positive step only
   local i = x
   if i <= y then
      while true do
         k(i)
         if y - step < i then
            break
         end
         i = i + step
      end
   end
end
```
be sufficient, or are there further corner cases?

mascarenhas:
The semantics is actually to do `i = i + step` with an overflow check, then test for the result of this check along with the regular `i <= y` test. A simple `for i = <start>, <limit>, <step>`  with a positive step goes from:

    do
      local i = <start>
      local _limit = <limit>
      local _step = <step>
      while i <= _step do
         <body>
         i = i + _step
      end
    end

to:

    do
      local i = <start>
      local _limit = <limit>
      local _step = <step>
      local _overflow = false
      while (not _overflow) and (i <= _step) do
         <body>
         i, _overflow = checked_add(i, _step)
      end
    end

where `checked_add` does the addition and also returns whether overflow occurred or not. This cover all the corner cases around overflow/underflow, while the alternative semantics you gave will break when `y - step` underflows due to a very small `y` or a very big `step`.

hugomg:
Another big consequence is that Titan for loops would have a different semantics than Lua for loops.

With respect to portability, it would also be harder to create a version of Titan that compiles down to Lua, if we ever wish do do something like that.

mascarenhas:
We only need a portable fallback for `__builtin_add_overflow` (introduced in GCC 5), which works like the `checked_add` that I used above. There are several portable options at https://blog.regehr.org/archives/1139, and MSVC also has similar operations like https://msdn.microsoft.com/en-us/library/windows/desktop/hh707046(v=vs.85).aspx.

mascarenhas:
Someone in the lua-l list seems to have found an alternative solution that we may check and compare with ours, although I suspect the overflow check will be cheaper, even if done in a fully portable way (with compiler intrinsics overflow checking is certainly faster). But the ongoing discussion also brought to attention the fact that the numeric for loop for floating point intervals has some nasty edge cases, too.

My gut tells me to simply restrict the numeric for loop to integers; getting a for loop to have nice closed interval semantics for any values assigned to start/finish/step is hopeless, I think their semantics can only be really understood operationally, as loop that does successive floating point additions and comparisons, so we might as well require the user to just write a `while` loop in that case... there are some bandaid fixes such as checking if `current+step == current` and raising a lack of precision error, but they do not fix the underlying problem.

hishamhm:
> My gut tells me to simply restrict the numeric for loop to integers; 

Having had used for loops with decimal steps many times over the years (in Lua and other languages) I think this is not something we can afford to discard.

> getting a for loop to have nice closed interval semantics for any values assigned to start/finish/step is hopeless, I think their semantics can only be really understood operationally

The alternatives become:

1. the simple thing to do is to set this issue aside and just live with Lua semantics;
2. do what Dirk Laurie suggests in lua-l and have two separate implementations of FP and integer for, each accounting for its specificities (but if FP `for` does the MATLAB thing and uses an epsilon, I'd expect FP `==` to use it as well — this is something I would find _very_ appealing for a scripting language aimed at domain specialists... but that's not the language _we_ are designing ;) ).
3. do what this patch does, and do the "right thing" for integers and just accept that FP numbers are weird

Pragmatically speaking, option 1 sounds the way to go for the time being.

mascarenhas:
> Pragmatically speaking, option 1 sounds the way to go for the time being.

Yes, that would be my choice too... it would be surprising to have nice closed interval semantics for integers, but while loop semantics for floating point, I think most people do not expect an innocent-looking `for <id>=<start>,<end>,<step> do` loop to be overloaded in this way.

> do what Dirk Laurie suggests in lua-l and have two separate implementations of FP and integer for, each accounting for its specificities (but if FP for does the MATLAB thing and uses an epsilon, I'd expect FP == to use it as well — this is something I would find very appealing for a scripting language aimed at domain specialists... but that's not the language we are designing ;) ).

In [Pyret](https://www.pyret.org/docs/latest/numbers.html) comparing two floating point numbers ("roughnums") for equality is an error, you have two use one of [a set of specialized operators](https://www.pyret.org/docs/latest/equality.html#%28part._s~3abounded-equalities%29) where you have to explicitly supply the tolerances.

hishamhm:
> In Pyret comparing two floating point numbers ("roughnums") for equality is an error, you have two use one of a set of specialized operators where you have to explicitly supply the tolerances.

A fine choice for a language aimed at teaching, as well!

PR #64, opened 2017-11-18
Add records to the parser

gligneul:
See the manual for more details about the initial proposal. This should be the bare minimum for implementing records. Initializer lists `{x = 10, y = 10}` can be implemented later.

### DIFF
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -36,7 +36,7 @@ types.Then = {
 
 types.Var = {
     Name    = {"name"},
-    Array   = {"exp1", "exp2"},
+    Bracket = {"exp1", "exp2"},
     Dot     = {"exp", "name"},
 }
 
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -35,7 +35,7 @@ types.Then = {
 
 types.Var = {
     Name    = {"name"},
-    Index   = {"exp1", "exp2"},
+    Array   = {"exp1", "exp2"},
 }
 
 types.Exp = {
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -4,7 +4,7 @@ local ast = {}
 local types = {}
 
 types.Type = {
-    Basic   = {"name"},
+    Name    = {"name"},
     Array   = {"subtype"},
 }
 
--- a/titan-compiler/ast.lua
+++ b/titan-compiler/ast.lua
@@ -11,6 +11,7 @@ types.Type = {
 types.TopLevel = {
     Func    = {"islocal", "name", "params", "rettype", "block"},
     Var     = {"islocal", "decl", "value"},
+    Record  = {"name", "fields"},
 }
 
 types.Decl = {
@@ -36,6 +37,7 @@ types.Then = {
 types.Var = {
     Name    = {"name"},
     Array   = {"exp1", "exp2"},
+    Dot     = {"exp", "name"},
 }
 
 types.Exp = {
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,3 +9,8 @@ insert_final_newline = true
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
+
+[*.titan]
+indent_style = space
+indent_size = 4
+trim_trailing_whitespace = true
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -1,6 +1,16 @@
 # Titan Language Reference
 
-## Arrays
+## Types
+
+### Basic Types
+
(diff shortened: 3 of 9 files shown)


titan-compiler/ast.lua
```diff
@@ -35,7 +36,8 @@ types.Then = {
 
 types.Var = {
     Name    = {"name"},
-    Index   = {"exp1", "exp2"},
+    Array   = {"exp1", "exp2"},
+    Dot     = {"exp", "name"},
```
hugomg: I prefer the original Var_Index name because it also applies to non-array things that we might eventually add (like maps).

hugomg (commented):
Looks good to me, modulo nitpicks. But I think it might be a good idea to try to implement initializer list parsing now just to be sure that they won't cause any problems at the syntax level (its OK if they fail during code generation)

hishamhm:
The `Index` vs `Array` vs `Dot` question actually brings an interesting question. Is dot-notation _not_ going to be syntax sugar for bracket notation?

Because bracket-notation for records means we get indirect addressing of record fields, which is a powerful (and desirable!) feature, but one which demands more from the type checker and code generator:

Given record type

```lua
record R
    x: A
    y: B
    z: C
end
```

and a record 

```
local r: R
```

as a coming-from-Lua programmer, I expect `r[s]` to be valid, have type `A | B | C`, and to perform the equivalent of the following pseudo-code:

```
if (s equals "x") {
   tmp = r.x;
} elseif (s equals "y") {
   tmp = r.y;
} elseif (s equals "z") {
   tmp = r.z;
} else {
   runtime error;
}
```

(Evidently, when `s` is a literal, I'd expect `r["x"]` to produce the same code as `r.x`)

This is an important feature because it will be our only way to do "pointers to record fields". This is probably beyond the scope of this PR (as it's probably best to implement dot-only records first then extend them later), but definitely something to keep in mind for anyone thinking about efficiency of records in Titan. :)

mascarenhas:
> as a coming-from-Lua programmer, I expect r[s] to be valid, have type A | B | C, and to perform the equivalent of the following pseudo-code:

If `s` is only known at runtime what you want is runtime reflection, and definitely something Titan should have (it can use the same mechanism that Lua uses to access record fields at first, and be as fast/slow as an access from Lua), but I am not sure we would like to allocate such prime syntactic real estate as `e1[e2]` to it, we might want to reserve this for future CTMP hooks.

> This is an important feature because it will be our only way to do "pointers to record fields". 

Actually having a type that is a mutable, typesafe reference to the innards of a Titan data structure (array, record, object) is rather straightforward, as everything is backed by a `TValue` anyway, and the Lua interpreter already has a builtin, GC-traceable value that is in essence a `TValue*` called `UpVal`. We can even be trendy and call these references "lenses". 😃

hishamhm:
> I am not sure we would like to allocate such prime syntactic real estate as e1[e2] to it, we might want to reserve this for future CTMP hooks.

`e1[e2]` is an archetypical syntax for indirect addressing, and is the one used in Lua to achieve that same goal. I'd like things that look the same to act the same whenever possible. CTMP will be a new feature and will demand new syntax. Overriding the syntax of such a basic Lua feature with a different one would be more confusing than anything.

So, my expectation would be that either `e1[e2]` is not supported in records at all or, if they do, they act the same as in Lua (modulo nullability behavior, since that's a general difference wrt typing in Titan).

titan-compiler/ast.lua
gligneul: I was thinking to rename them to `Var_BracketIndex` and `Var_DotIndex` or the less verbose version `Var_Braket` and `Var_Dot`.

mascarenhas:
Having records be Titan's table-analogs and `e1[e2]` be runtime reflection for them is reasonable, but I think too constraining for objects (think userdata, and `e1[e2]` being whatever the userdata decides it is).
It is still an open question on whether objects and classes will subsume records later on, or they will both coexist.

AFAIK Lua and Javascript the only language where `[]` is runtime reflection for object-like data structures. For both Ruby and Python `[]` is an overridable method call on the object, mimicking the behavior of statically-typed languages where there is operator overloading for `[]`.

titan-compiler/ast.lua
mascarenhas: +1 to `Var_Bracket` and `Var_Dot`, with corresponding `Exp_Bracket` and `Exp_Dot` for the rvalue versions.

gligneul: Do we need the `Exp_Bracket` and `Exp_Dot`? At the time we are using `Exp_Var` + `Var_*` for everything.

hugomg (approved):

PR #66, opened 2017-11-23
more work on modules

mascarenhas:
I have done some refactoring of the typechecker to have a single point where we are reading from/assigning to something that is not a first-class value. This should be merged into master before continuing work on records.

Typechecking for modules is done, BTW, and I can move to code generation and documentation. Once I get modules working I am going to refactor the compilation pipeline to work on "projects" (a set of modules and their transitive imports) instead of single source files.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -54,3 +54,85 @@ You can access the fields with the dot operator:
     local a = p.x
     p.y = 2
 
+## Modules
+
+A Titan source file (with a `.titan` extension) is a *Titan module*. A Titan module
+is made-up of *import statements*, *module variable declarations*, and *module function
+declarations*. These can appear in any order, but the Titan compiler reorders
+them so all import statements come first (in the order they appear), followed by
+variable declarations (again in the order they appear), and finally by function
+declarations, so an imported module can be used anywhere in a module,
+an module variable can be used in any function as well as variable declarations
+that follow it, and module functions can be used in any other function.
+
+### `import` statements
+
+An `import` statement references another module and lets the current
+module use its exported module variables and functions. Its syntax
+is:
+
+    local <localname> = import "<modname>"
+
+The module name `<modname>` is a name like `foo`, `foo.bar`, or `foo.bar.baz`.
+The Titan compiler translates a module name into a file name by converting
+all dots to path separators, and then appending `.titan`, so the above three
+modules will correspond to `foo.titan`, `foo/bar.titan`, and `foo/bar/baz.titan`.
+All paths are relative the the path where you are running `titanc`.
+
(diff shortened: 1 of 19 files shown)


mascarenhas:
Ok, I did a pretty good start on code generation, have to add machinery in `coder_spec.lua` to write tests for it (for now changed `titanc` to be module-aware, moved some functions from `artisanal.titan` to a separate `utils.titan` module and manually tested calling these functions through `artisanal.titan`).

mascarenhas:
By the way, Titan -> Titan module loading is *not* done by `dlopen`, but by the operating system loader itself. But this is easily changeable in the future, if we want greater control on where Titan modules find their dependencies at runtime.

hishamhm:
Oh, I think the renaming of the `lua` binary is quite controversial (and IMO technically incorrect: we'd be in a weird situation where a program file starting with `#!/usr/bin/env titan` _isn't_ Titan code). I understand the convenience benefit of the custom-built Lua interpreter, and the increased robustness of avoiding mixups with system-installed Lua, but especially in a mixed-language environment, we shouldn't mix their names like that. 

Need to think of a different solution, ideally outside of this PR.

mascarenhas:
I can rename it back to `lua`, but I do not know if it is more controversial to rename the interpreter or just replace the `lua` interpreter of the user's system with our version that can load Titan modules...

It is not just the exported symbols, I also added a `/usr/local/lib/titan` to `package.cpath` so the user's Lua interpreter will not try to load Titan modules by accident.

mascarenhas:
I have renamed the interpreter back to `lua` to not block this merge. I have not figured out a good way for Titan modules to detect that they are being loaded by the correct version of Lua, as `_VERSION` is `Lua 5.3` and not `Lua 5.3.4`. If we change `luaconf.h` to make `_VERSION` be `Lua 5.3.4` it might break other Lua modules, and there is no way to add a new `_RELEASE` variable through `luaconf.h`, we need to patch other files.

doc/manual.md
```diff
@@ -54,3 +54,85 @@ You can access the fields with the dot operator:
...
 
+## Modules
+
+A Titan source file (with a `.titan` extension) is a *Titan module*. A Titan module
+is made-up of *import statements*, *module variable declarations*, and *module function
+declarations*. These can appear in any order, but the Titan compiler reorders
+them so all import statements come first (in the order they appear), followed by
+variable declarations (again in the order they appear), and finally by function
+declarations, so an imported module can be used anywhere in a module,
+an module variable can be used in any function as well as variable declarations
```
hishamhm, line 65: `an module` -> `a module`

hishamhm (approved):

PR #67, opened 2017-11-24
Minor cleanups

hishamhm:


titan-compiler/coder.lua
```diff
@@ -1327,10 +1326,6 @@ int $LUAOPEN_NAME(lua_State *L) {
 }
 ]]
 
-local initsig = [[
-    void $INITNAME(lua_State *L);
-]]
```
hishamhm, line 1332: @mascarenhas this unused variable might be relevant

mascarenhas: It is not, it is a remnant of when I was generating header files instead of just inlining external module declarations.

mascarenhas (approved):

PR #68, opened 2017-11-24
Move EBNF inside manual.md

andremm:
In this PR I propose to move the EBNF file inside the `manual.md`.  I believe this way it will be easier to keep track whether we are updating our documentation of the Titan syntax whenever we change the parser.


hishamhm (approved):
Good idea!

PR #69, opened 2017-11-28
Improving error messages

mascarenhas:


titan-compiler/syntax_errors.lua
```diff
@@ -26,6 +27,108 @@ local errors = {
...
+        msg = "Expected 'end' to close the record." },
+    
+    { label = "FieldRecord",
+        msg = "Expected a field in record declaration." },
+    
+    { label = "NameImport",
+        msg = "Expected a name after 'local'." },
+   
+   --this label is not thrown in rule 'import' because rule 'toplevelvar'
+   --matches an invalid input like "local bola import"
```
hugomg, line 68: I'm confused about this commend. The ASSIGN in the toplevelvar rule is not optional so it shouldn't be able to parse "local bola import" right?

titan-compiler/syntax_errors.lua
```diff
@@ -26,6 +27,108 @@ local errors = {
...
+   { label = "AssignImport",
+        msg = "Expected '='." },
+   
+   --this label is not thrown in rule 'import' because rule 'toplevelvar'
+   --matches an input like "local bola = X", given that X is a valid expression,
+   --or throws a label when X is not a valid expression
+   { label = "ImportImport",
+        msg = "Expected 'import' keyword." },
```
hugomg, line 76: Perhaps a way to avoid this issue would be to syntactically allow assignment statements at the toplevel and use a `=>` capture to tell which of them are import declarations, which are toplevel variables and which should be considered syntax errors.

hugomg (commented):
This looks very nice overall. The only thing bugging me is that now I think I should have inlined all the test cases inside the spec file instead of having separate files for them...

mascarenhas:
@hugomg, can you approve and merge this so work on the parser can continue?

hugomg (commented):
Looks good!

hugomg (approved):

PR #72, opened 2017-11-30
Syntactic support for higher order functions

hugomg:
This patch adds support for function types in Titan's lexer and parser. We should be able to merge this now but we could also wait until the code generation is done, if preferred.

Function types currently require parentheses around both the list of argument types and return types, as in Typed Lua. I tried to think of a way to allow omitting the parenthesis when the list has exactly one element but I couldn't figure out a way to do that while also forbidding weird things like `(a , b -> c, d)`. If anyone can think of an elegant solution to this I am all ears. (Btw, I checked how Typescript approaches this and they also have mandatory parenthesis on the argument types. The return type doesn't have parentheses because it is always one value)


spec/parser_spec.lua
```diff
@@ -91,6 +91,44 @@ describe("Titan parser", function()
...
+
+            -- multiple input parameters
+            { decl = { type = {
+                _tag = "Type_Function",
+                argtypes = {
+                    { _tag = "Type_Name", name = "a" },
+                    { _tag = "Type_Name", name = "b" } },
+                rettypes = { { _tag = "Type_Name", name = "c" } } } } },
+
+            -- right associativity of ->
```
gligneul, line 115: Do we have associativity in `->`? If not I would rather remove this commentary.

gligneul (commented):
Looks great!

mascarenhas:
Why would you want to forbid `(a , b -> c, d)`? This is the parameter type list for a function that has one parameter of type `a`, one that is a function from `b` to `c`, and a third parameter that is a `d`. It is completely unambiguous.

hugomg:
I think a context free grammar for types with optional parenthesis would look like this:

```
simpletype
    = NIL
    | NAME
    | '{' type '}'

type
    = simpletype
    | types '->' types

types
    = type
    | '(' comma_separated_type_list ')'

comma_separated_type_list
    = /*empty*/
    | type
    | type ',' comma_separated_type_list
```

However, to convert this to an unambiguous LPeg grammar we would need to make the '->' operator right associative and get rid of the indirect left recursion. I am having a hard time figuring out how to do that. Can anyone help me here?

mascarenhas:
> I am having a hard time figuring out how to do that. Can anyone help me here?

It is kind of like any other operator:

```
type -> '(' types? ')' '->' type / simpletype ('->' type)?
simpletype -> NIL / NAME / '{' type '}'
types -> type (',' type)*
```

By the way, if `nil` is a keyword following Lua why are the names of other builtin types keywords too (`integer`, `float`, `boolean`, `string`)? Typed Lua couldn't add new keywords, but Titan does not have this handicap.

hugomg:
I think we should be able to omit the parentheses now and the comment on the test suite about right associativity makes sense again :)

Along the way I got rid of the "expected -> after )" error label to simplify the grammar a little.

doc/manual.md
```diff
@@ -28,6 +28,31 @@ The only time where no context is available is when you declaring a new
...
+example, `(a, b) -> (c)` is the function type for a function that receives one
+argument of type `a` and one of type `b` and returns a value of type `c`. For
+function types that only receive one input parameter or return a single value,
+the parentheses are optional. For example, `int -> string` is the type of
+functions that map `int` to `string`. The `->` type constructor is right
+associative. That is, `a -> b -> c` is equivalent to `a -> (b -> c)`, the type
+of functions that receive an `a` and return a function that receives a `b` and
+return a `c`.
+
+A titan variable of function type may refer to either statically-typed Titan
```
hishamhm, line 43: typo: titan → Titan

hishamhm (approved):
Apart from the minor typo, LGTM! I say let's merge this and then make code generation a second PR.

PR #73, opened 2017-12-04
Inline parser tests

hugomg:
I reviewed the lexer and parser test suite and cleaned it up a bit by putting all the test files inside the spec file itself, which is more readable. The original reason for separating the test files was because `assert.are.same` didn't give helpful error messages for errors more than 3 or 4 levels deeps in the AST so we relied on titanc's --print-ast for that. But now that I changed the test cases to call assert.are.same deeper down in the AST instead of always calling it at the root this is not an issue anymore. One side bonus of this is that now the test can contain just the type/statement/expression we are testing, without also needing to put it inside a function in the toplevel.

Along the way I also discovered some bugs that had been hanging around these parts. I created new (pending) test cases for those and going to open separate github issues for each.


gligneul (approved):
Impressive work, the tests are way more legible now (+1 for inlining the testfiles).
Also, I love when commits have more removed lines than added ones. :)

PR #80, opened 2017-12-04
Add initializer lists to the parser

gligneul:
This is a proposal to change the syntax to create records.
This also have a impact on arrays, since `local x = {}` will be an error instead of an array of integers.
See the modifications in the manual for more details.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -22,9 +22,11 @@ For example, if you are assigning the new array to an array-typed variable, the
 array will have the same type of the variable.
 If you are passing the array as an argument to a function that expects an
 array-type parameter, the new array will have the same type as the parameter.
-When declaring a new variable, you should explicitly specify the type of the
-array; `local a = {}` will yield a compile time error.
-
+If you are declaring a new variable with an explicit array type declaration,
+the new array will have the type you declared.
+The only time where no context is available is when you declaring a new
+variable and you have not given a type to it; in that case the array will have
+type `{ integer }`.
 
 ### Functions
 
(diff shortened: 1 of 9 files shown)


doc/manual.md
```diff
@@ -55,29 +53,27 @@ return values for functions.
...
-    local c = Circle.new(p, 3.5)
-
-You can access the fields with the dot operator:
-
-    local a = p.x
-    p.y = 2
+You can create records with initializer lists, assigning a value to each field
+either by it's name or positional order.
+For instance, you could initalize an instance of the record `Point` with: `{ x =
+3, y = 5 }`, `{ y = 5, x = 3 }` or `{ 3, 5 }`.
```
hishamhm: :-1: to `{ 3, 5 }` — very misleading!

hugomg: I think it is important that it be possible to instantiate records without needing to type all the field names. I don't have a strong preference between a Lua-like `local p = Point.new(3,5)` or a C-like `local p : Point = {3, 5}` though. Just don't make me write `local p : Point = Point.new(3,5)`  :)

doc/manual.md
```diff
@@ -55,29 +53,27 @@ return values for functions.
...
+In all those cases, the field `x` will receive the value `3` and the field `y`,
+`5`.
+Like arrays, Titan will try to infer the type of the initializer list, but it is
+necessary to specify the type of the record when declaring a new variable
+(`local p: Point = { 3, 5 }`).
+
```
hugomg: I don't know if this is out of scope for this PR but currently our type inference rules are undocumented and untested. I think we should rectify that if we are going to strongly rely on type inference for record instantiation.

doc/manual.md
mascarenhas: I agree with @hishamhm . If you want positional having `local p = Point.new(3,5)` is good enough for me (and supported by Titan's limited type inference). Having `local p: Point = { 3, 5 }` opens up the question of what things like `local p: Point = { x = 3, 5 }`, `local p: Point = { y = 5, 3 }`, or `local p: Point = { 5, x = 3 }` mean...

gligneul: Removed the `{ 3, 5 }` syntax for records. Now we have different ast nodes for arrays constructors and records initalizer lists, so things like `{ y = 5, 3 }` are syntax errors.

doc/manual.md
gligneul: I don't know how to express that without being repetitive (there is already an explanation on the array section). Maybe we should have a section just for the inference rules, but that is indeed beyond the scope of this PR.

hugomg (changes requested):
Turning `{x=11,  22}` into a syntax error will stop Titan from creating plain Lua tables with mixed array and hash parts. We currently don't implement this feature but I think it would be very desirable in the future for Lua interoperability.

I think the type checker would be a better place to catch these array and record initialization errors.

mascarenhas:
> Turning {x=11, 22} into a syntax error will stop Titan from creating plain Lua tables with mixed array and hash parts. We currently don't implement this feature but I think it would be very desirable in the future for Lua interoperability.

Just because it is a syntax error now does not mean it must be always a syntax error! If it is easy enough to forbid in the parser I do not see a need to make the typechecker more complex right now for the sake of a future feature.

mascarenhas (approved):

gligneul:
> Just because it is a syntax error now does not mean it must be always a syntax error! If it is easy enough to forbid in the parser I do not see a need to make the typechecker more complex right now for the sake of a future feature.

Even tough I agree with you, the changes in the typechecker might be simpler than the changes I've made to the parser. Don't merge it yet, let me check first please.

hugomg (approved):

PR #81, opened 2017-12-04
Load and link Titan modules with dlopen and dlsym

mascarenhas:
Load Titan modules with dlopen so we can have better control of where the modules come from. Keeps a cache of loaded module handles in the registry, similar to what Lua does. The default module search path is  `/usr/local/lib/titan/0.5;.` (should we reverse that), but a `TITAN_PATH` environment variable overrides that. The search path is a list of semicolon-separated paths, and not a "path template" like in Lua.


hishamhm:
I'm uncertain about supporting versioned environment variables. In my experience with Lua, they've been more troublesome than helpful (especially when there's an unversioned fallback). I need to give this subject a little more thought.

mascarenhas:
> I'm uncertain about supporting versioned environment variables. In my experience with Lua, they've been more troublesome than helpful (especially when there's an unversioned fallback). I need to give this subject a little more thought.

I followed Lua's lead, but my suggestion would be to remove the unversioned fallback...

By the way, I have implemented your suggestion for letting the compiler load serialized type information from a compiled module. Right now the compiler will always look for a compiled module first, and only look for a source module if it fails, but I am open to other default policies, like looking for a local compiled module, then for a local source module, and finally for a system-wide compiled module.

mascarenhas:
I do not condone that, but the possibility of loading type information from a shared library is also a backdoor for a "poor-man's FFI", where we can write C code that pretends it is a Titan module and is callable by other Titan modules.

But I really want to start working on a way for Titan code to interface with C code using as input something like what @hishamhm's header parser will output, so we can write the standard library in Titan itself.

hugomg:
I have to confess that I will need some more time to be able to fully understand how our modules are working. The part I could understand looks fine and nothing bad jumped at my eyes.

mascarenhas:
> Right now the compiler will always look for a compiled module first, and only look for a source module if it fails, but I am open to other default policies, like looking for a local compiled module, then for a local source module, and finally for a system-wide compiled module.

Another possible default policy would be to do as @hugomg suggested once and make `titanc` only compile a single module at a time. The compiler would only look for compiled modules, and it would be the programmer's responsibility to use a build system such as `make` to keep everything in sync.

mascarenhas:
I have tweaked the policy to always recompile the module if its source is newer than the binary, even if a binary is present, and changed `titanc` to take module names and make the path to the sources a command-line option. Can we merge this and decide on what policy we want to keep later?

titan-compiler/coder.lua
```diff
@@ -1305,6 +1311,130 @@ $INCLUDES
 
 ]]
 
+local libopen = [[
```
gligneul, line 1315: We could store this the preamble in a separate file.

mascarenhas: The problem is where to load it from when running the compiler. Remember that `coder.lua` is a regular Lua module, loaded by `require`. We would probably need another configurable path for the compiler, which is too much of a hassle until our "runtime" grows more.

gligneul: Yes, this don't need to be done now, I'll create a issue deal with the runtime.

andremm (approved):

PR #82, opened 2017-12-04
check for malformed decimal escapes, fixes #75

### LINKED ISSUE #75: Large decimal string escapes crash the compiler
If a three-digit decimal string escape is too large (greater than 255) it will crash the compiler with an uncaught exception (bad argument to string.char). For example, the following program crashes titanc when we try to compile it

    function f(): string
        return "\555"
    end

Ideally we should detect this and raise an `MalformedEscape_decimal` parsing error. This would be super easy to do if lpeglabel implemented this [feature request](https://github.com/sqmedeiros/lpeglabel/issues/19) of mine but I think it should also be possible to implement without that.

mascarenhas:


titan-compiler/lexer.lua
```diff
@@ -95,6 +95,14 @@ do
...
             end)
 
+    local decimal_escape = P("1") * R("09") * R("09") +
+        P("2") * R("04") * R("09") * -R("09") +
+        P("2") * P("5") * R("05") * -R("09") +
+        P("0") * R("09") * R("09") * -R("09") +
+        R("09") * R("09") * -R("09") +
+        R("09") * -R("09") +
+        R("09") * T(labels.MalformedEscape_decimal)
+
```
hugomg: I think we need to get rid of the `-R("09")`. This code will reject things like `\"2490`, which is should accept ("\249" followed by "0").

hugomg (changes requested):
I would rather solve this with a match-time capture. I think a `0 <= n and n <= 255` written in Lua would be clearer than a clever regular expression.

titan-compiler/lexer.lua
mascarenhas: I guess this is in the eye of the beholder, I am not fond of overusing math-time captures myself... will fix the second point.

hugomg (approved):

PR #90, opened 2017-12-07
 Initial support for multiple function returns

hugomg:
The types.Function datatype now has a list of return types instead of a single return type. This updates it to follow the current parser behavior and the future direction for the language.

Multiple returns are not yet implemented though. An attempt to declare a function with multiple return types will crash the compiler with a "not implemented" type error.


mascarenhas (approved):
Looks good. I think the next step is to make the default return type be an empty type list instead of a list containing nil (and a `return nil` would be a compile-time error inside a function with this type).

PR #92, opened 2017-12-08
Add 'value' type and explicit casts

mascarenhas:
The `value` type is Titan's dynamic type (in the gradual typing sense). Any type can be cast to `value`, and `value` can be cast to any type (the "downcast" is checked at runtime). 

If a `value` is expected but a term with any other type is supplied a coercion to `value` is inserted, and there are several places where coercions from `value` to some other type can also be inserted. For example, trying to add two `value`s is a type error, but adding a `value` to an integer will insert a cast that will try to get an integer from the `value` at runtime.

This PR also adds a `<simpleexp> as <type>`  explicit cast/coercion.

Next step is to write new tests for the parser and typechecker to check the new code paths, and add the code to the typechecker for dealing with `value` and its coercions, as well as tests for this.


titan-compiler/types.lua
```diff
@@ -36,7 +36,54 @@ function types.has_tag(t, name)
...
+            types.equals(target, types.Integer)) or
+           (types.equals(target, types.Boolean) and
+            not types.equals(source, types.Boolean)) or
+           (types.equals(target, types.Value) and
+            not types.equals(source, types.Value)) or
+           (types.equals(source, types.Value) and
+            not types.equals(target, types.Value))
+end
+
+function types.compatible(t1, t2)
```
hugomg, line 55: Is this the gradual consistency relation?

titan-compiler/types.lua
```diff
@@ -36,7 +36,54 @@ function types.has_tag(t, name)
 end
 
 function types.is_gc(t)
-    return t._tag == "String" or t._tag == "Array"
+    return t._tag == "String" or t._tag == "Array" or t._tag == "Value"
+end
+
+function types.coerceable(source, target)
```
hugomg, line 42: Could you clarify what this function is doing (with a comment)? One thing I didn't quite understand here is if casting a type to itself (`intvar as integer`) is an error. I think adding some tests for the `as` expressions would help clarify this.

hugomg (commented):
I think adding a type that can pass around dynamic Lua values is something that we would have to do sooner or later and I feel that this is going in the right direction. That said, the specifics of how this is working now are not totally clear to me yet. Is the `value` behaving more like Top/Object, like C#'s `dynamic` or like Gradual Typing's `any`? I'm ambivalent about whether we should wait this PR now or wait until we add some extra tests or documentation to clear these things up.

In the spirit of unifying all the different type coercion operators, I also wonder: would it make sense to move the type testing from [checkandget](https://github.com/titan-lang/titan/blob/665c5ebfd3d6419c6b104c2892b5a95bf66e2b7e/titan-compiler/coder.lua#L74) and [checkandset](https://github.com/titan-lang/titan/blob/665c5ebfd3d6419c6b104c2892b5a95bf66e2b7e/titan-compiler/coder.lua#L114) inside the generated code for an Exp_Cast expression?

titan-compiler/types.lua
mascarenhas: This function checks whether we can coerce source to target, even though they are different types, to guide the automatic insertion of casts. About no-op casts, currently they are not an error (and I plan to make them be a no-op in the code generator).

titan-compiler/types.lua
mascarenhas: If you are talking about `types.compatible`, yes, it is gradual consistency.

mascarenhas:
> Is the value behaving more like Top/Object, like C#'s dynamic or like Gradual Typing's any? 

It is more like gradual typing's `any` or `*`, in the compiler tries to insert automatic coercions from it to other types where appropriate, but apart from these coercions does not generate different code from what is currently generating (which would be the case for a C# `dynamic`, or even Typed Lua's `any`).

> In the spirit of unifying all the different type coercion operators, I also wonder: would it make sense to move the type testing from checkandget and checkandset inside the generated code for an Exp_Cast expression?

Not with the current state of our back-end, as it would just make the code have extra copies of `Value`s, and even pushing things to the stack twice. In fact what I will actually do is the opposite: an `Exp_Cast` where the expression is `Var_Bracket` will actually delegate the actual cast to `Var_Bracket`, so we don't copy the `Value` in the slot unecessarily.

mascarenhas:
Still need to write docs for this, but if anyone wants to ok and merge it so others can check for conflicts I am ok with it.

mascarenhas:
Will write docs ASAP, but removed automatic coercions from `value` to other types in unary and binary operations (except for `==` and `~=`), where it might lead to "raw" semantics by accident (I am keeping the semantics of `==` and `~=` raw for now). Now using `value` in any binary/unary operation except for `==`, `~=`, `and`, `or`, and `not` is a type error (the last three are justified by the fact that they do not have metamethods in Lua, so are already always raw).

If the programmer wants to add a `value` to an `integer`, or compare a `value` to a `string` with `<`, or do bitwise negation on a `value`, now she has to explicitly cast the `value` to something that Titan knows hows to compile. This should make @hugomg and Roberto happier. 😄

mascarenhas:
Is someone willing to ok and merge this? 😄

andremm (approved):

andremm:
> Is someone willing to ok and merge this? 😄 

OK, I can do that. :)

PR #94, opened 2017-12-12
Guidelines to ensure our C code indents well

### LINKED ISSUE #93: Pretty-printer doesn't understand comments
The C files our compiler is currently generating are terribly indented because it is not recognizing open braces that are followed by comments. In cases such as the following, it continues at the same indentation level instead of increasing the indentation level by one.

    if (path == NULL) { /* no environment variable? */

A possible workaround would be to change our style guide to require that comments appear in lines by themselves. A more technical solution would be to make the reindent_c function a bit more clever.

What do you think would be the best course of action here?

hugomg:
Fixes #93

### DIFF
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -590,8 +590,10 @@ local function codeassignment(ctx, node)
                 } else {
                     _slot = (TValue *)luaH_getint(_t, _k);
                     TValue _vk; setivalue(&_vk, _k);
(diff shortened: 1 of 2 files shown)


mascarenhas (approved):

mascarenhas:
Can I merge this, or will you want to do more commits?

hugomg:
Go ahead. I'm not planning on adding more commits to this branch.

PR #96, opened 2017-12-13
Higher order functions work in progress

hugomg:
This is my work in progress branch for implementing higher order functions. Perhaps it might be a good idea to merge it now before it diverges too much from master... The most notable things so far are:

- Got rid of code that forbid using functions as variables (but didn't actually implement the higher order functions yet)
- Removed some accidentally duplicated tests from checker_spec. Those probably got introduced during a git merge a while back.
- Forbid assigning to global variables defined in other modules. Function variables introduces some confusion about what is a constant toplevel function and what is a mutable global variable containing a function reference. For now, I think it was easier to just make all external global variables off-limits.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -279,20 +279,6 @@ describe("Titan type checker", function()
         assert.truthy(ok)
     end)
 
-    it("type-checks 'for'", function()
-        local code = [[
-            function fn(x: integer): integer
-                local i: integer = 0
-                for i = 1, 10 do
-                    x = x + i
-                end
-                return x
-            end
-        ]]
-        local ok, err = run_checker(code)
-        assert.truthy(ok)
-    end)
-
     it("type-checks 'while'", function()
         local code = [[
             function fn(x: integer): integer
@@ -888,6 +874,7 @@ describe("Titan type checker", function()
             end
         end
     end
+
     it("returns the type of the module with exported members", function()
         local code = [[
             a: integer = 1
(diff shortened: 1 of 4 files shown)


mascarenhas:
> Forbid assigning to global variables defined in other modules. Function variables introduces some confusion about what is a constant toplevel function and what is a mutable global variable containing a function reference. For now, I think it was easier to just make all external global variables off-limits.

Please revert this. Variables should be variables, period. It is trivial to forbid assignment to module functions, even if they have the same type as module variables that happen to have functions in them. There is no confusion, it is just like C with functions versus variables that have function pointer types, where you can call both, but only use one of them in the left side of an assignment.

Remember the mantra: Titan is not Lua, function declarations are not syntax sugar for assignments, the toplevel is not imperative.

hugomg:
Allowing mutable variables again isn't just a matter of reverting this patch. At the moment the type of a module (as created by types.maketype) only carries information about the type of an identifier and cannot distinguish between toplevel identifiers that are mutable function variables and toplevel identifiers that are constant functions. To fix that we would need to change this datatype to include the necessary extra information somehow or we would need to extend our type system with some sort of const qualifier or something else along these lines. I just didn't want to deal with this for now.

I never intended to permanently get rid of assignment to external globals. This is why the related tests are marked as "pending" instead of just being deleted.

mascarenhas:
> extend our type system with some sort of const qualifier

It is not constness, with the introduction of first-class functions module variables and functions should be separate sets instead of just a `members` set like it currently is. The expression `foo.bar` has different codegen if `bar` is a variable vs a module function, even when no assignment is involved.

hugomg:
I mentioned constness because I expect that in a world with constant variables the compiler would be able to produce the same code for toplevel functions (as we have today) and constant function variables.

gligneul:
> I mentioned constness because I expect that in a world with constant variables the compiler would be able to produce the same code for toplevel functions (as we have today) and constant function variables.

Even if a function pointer is constant, the compiler might not be able to resolve it at compile time like top level functions, so the generated code will be different. I think we need a separate type for top level functions and first class functions.

PR #97, opened 2017-12-17
Refactor the typecheck of toplevel nodes

gligneul:
We have now only one pass that typechecks all toplevel nodes. This way the duplicate code could be merged together and new top level declarations can be easily added.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -92,7 +92,7 @@ describe("Titan type checker", function()
         ]]
         local ok, err = run_checker(code)
         assert.falsy(ok)
-        assert.match("duplicate function", err)
+        assert.match("duplicate declaration", err)
     end)
 
     it("catches duplicate variable declarations", function()
@@ -108,7 +108,7 @@ describe("Titan type checker", function()
         for _, c in ipairs(code) do
             local ok, err = run_checker(c)
             assert.falsy(ok)
-            assert.match("duplicate variable", err)
+            assert.match("duplicate declaration", err)
         end
     end)
 
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -112,6 +112,55 @@ describe("Titan type checker", function()
         end
     end)
 
+
+    it("allows constant variable initialization", function()
+        local code = [[
+            x1 = nil
+            x2 = false
+            x3 = 11
+            x4 = 1.1
+            x5 = "11"
+            x6 = {}
+            x7 = {1, 2}
+            x8 = "a" .. x5
+            x9 = 1 + x3
+            x10 = not x2
+            x11: integer = 10.1
+        ]]
+        local ok = run_checker(code)
+        assert.truthy(ok)
+    end)
+
+    it("catches non constant variable initialization in top level", function()
+        local code = {[[
+            function f(): integer return 10 end
+            x = f()
+        ]], [[
+            function f(): integer return 10 end
+            x = -f()
+        ]], [[
+            function f(): integer return 10 end
+            x = 10 + f()
+        ]], [[
+            function f(): integer return 10 end
+            x = "a" .. f()
+        ]], [[
+            function f(): integer return 10 end
+            x: float = f()
+        ]], [[
+            function f(): integer return 10 end
+            x = {f()}
+        ]], [[
+            x = {10}
+            y = x[2]
+        ]]}
+        for _, c in ipairs(code) do
+            local ok, err = run_checker(c)
+            assert.falsy(ok)
+            assert.match("must be constant", err)
+        end
+    end)
+
     it("catches variable not declared", function()
         local code = [[
             function fn()
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -2,6 +2,7 @@ local checker = require 'titan-compiler.checker'
 local parser = require 'titan-compiler.parser'
 local types = require 'titan-compiler.types'
 local driver = require 'titan-compiler.driver'
+local util = require 'titan-compiler.util'
 
 local function run_checker(code)
     driver.imported = {}
@@ -32,6 +33,17 @@ local function restrict(t1, t2)
     else
         return t2
     end
+end
+
+local function assert_type_check(code)
+    ok = run_checker(code)
+    assert.truthy(ok)
+end
+
+local function assert_type_error(expected, code)
+    local ok, err = run_checker(code)
+    assert.falsy(ok)
+    assert.match(expected, err)
 end
 
 -- To avoid having these tests break all the time when we make insignificant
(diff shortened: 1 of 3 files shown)


titan-compiler/checker.lua
```diff
@@ -689,43 +621,103 @@ local function checkfunc(node, st, errors)
...
+        return node.decl.name
+    elseif tag == "TopLevel_Func" then
+        return node.name
+    else
+        error("tag not found " .. tag)
+    end
+end
+
+-- Typecheck the toplevel node
+local tlcheckers = {
```
hugomg, line 695: I think that an if-else-if for this kind of control flow would be more consistent with the rest of the program than this tlcheckers table. (I think the table driven approach would fit better if we were using a visitor function for iteration)

titan-compiler/checker.lua
```diff
@@ -689,43 +621,103 @@ local function checkfunc(node, st, errors)
...
+}
+
+-- Colect type information of toplevel nodes
+--   ast: AST for the whole module
+--   st: symbol table
+--   errors: list of compile-time errors
+--   annotates the top-level nodes with their types in a "_type" field
+--   annotates whether a top-level declaration is duplicated with a "_ignore"
+--   field
+local function checktoplevel(ast, st, errors, loader)
```
hugomg, line 753: I am not sure if this is the exact same behavior as before. IIRC, the old version behaved like Javascript, effectively hoisting function variable declarations to the top of the module but keeping everything else the same while in this new version all the variable declarations are hoisted, as if there were a big letrec for the toplevel. Am I getting this right?

(In any case, I guess this special toplevel scoping is something we could discuss again)

titan-compiler/checker.lua
```diff
@@ -740,18 +732,21 @@ end
...
+--
+--   returns true if typechecking succeeds, or false and a list of type errors
+--   found
 --   annotates the AST with the types of its terms in "_type" fields
 --   annotates duplicate top-level declarations with a "_ignore" boolean field
---   loader: the module loader, a function from module name to its AST, code, and filename or nil and an error
 function checker.check(modname, ast, subject, filename, loader)
-    loader = loader or function () return nil, "you must pass a loder to import modules" end
+    loader = loader or function ()
+        return nil, "you must pass a loder to import modules"
```
hugomg: Did we already tell Hisham that we currently use a *lode* to *run* our modules? :)

Perhaps we should take this opportunity to fix this old typo here.

titan-compiler/checker.lua
```diff
@@ -689,43 +621,103 @@ local function checkfunc(node, st, errors)
...
+    local tag = node._tag
+    if tag == "TopLevel_Import" then
+        return node.localname
+    elseif tag == "TopLevel_Var" then
+        return node.decl.name
+    elseif tag == "TopLevel_Func" then
+        return node.name
+    else
+        error("tag not found " .. tag)
+    end
```
hugomg, line 691: This kind of thing is why I tried to make everything in the toplevel into a Decl. I gave up because that ended up breaking too much stuff but I still wonder if it would be a good way to simplify things in the future.

hugomg:
I do appreciate how this puts the name resolution in a single place. It is easier to understand than before.

titan-compiler/checker.lua
gligneul: I don't know about they being a Decl, but all of them should have a `name` field, so this function wouldn't be necessary. I will leave this modification for the future though.

titan-compiler/checker.lua
gligneul: I think the whole typechecker should follow this pattern. It is easier to merge common parts in the code.

hishamhm: I'm a fan of more table-based (in a Titan-future object-based?) approach like this rather than long sequences of if-else-if. And it's something we can transition to gradually (even if it results in less "consistency" in the meantime). I'm a fan of incremental improvement!

titan-compiler/checker.lua
hishamhm: > I am not sure if this is the exact same behavior as before.

A regression test should be able clear this question! Can you express your concern as a short test case?

titan-compiler/checker.lua
hishamhm: hahaha, or rename the variable to `loderunner` 😆

titan-compiler/checker.lua
gligneul: The older version sorted the declarations by: imports, variables, functions and then function bodies. This led to strange error messages:
```
function foo() end
foo = 0
```
This program had a duplicate declaration error in line 1, not line 2.
Is there any advantage in the old behavior that I'm not seeing?

titan-compiler/checker.lua
gligneul: Done :)

mascarenhas:
I am not against having more simpler top-level scoping rules, so as to avoid edge cases such as the double declaration that @gligneul mentioned, as long as we can still declare functions in any order. Calling functions of the same module in the initialization expressions of module variables should also be forbidden, as this would make it possible to access unitialized variables, like the example below shows (I should add this as a test case):

```
local function foo(): integer
  return bar()
end

local a: integer = foo()

local function bar(): integer
  return a
end
```

I have not ran it, but it looks like this typechecks with the code in this PR.

A more complex but consistent rule that might make @hugomg happier is to not treat the whole set of functions as a `letrec`, but only adjacent functions, so the above example would not typecheck because `bar` would not be visible in the body of `foo`, but changing the body of `foo` to `return 2` would make it typecheck. The following example would also typecheck correctly because `foo` and `bar` would be visible in the bodies of both functions:

```
local function foo(): integer
  return bar()
end

local function bar(): integer
  return 2
end

local a = foo()
```

hishamhm:
> A more complex but consistent rule that might make @hugomg happier is to not treat the whole set of functions as a letrec, but only adjacent functions

I had to read this and look at the example 3 times until I understood what you meant by that. So this proposal definitely falls in the "difficult to explain" category to me. :)

What is the problem with the `letrec`-style definition for functions again? I think it's pretty sensible to allow only constant expressions in module variable definitions.

gligneul:
> What is the problem with the letrec-style definition for functions again? I think it's pretty sensible to allow only constant expressions in module variable definitions.

I also think this is reasonable.

hugomg:
One downside of the global letrec is that it is incompatible with imperative code, which rules out adding an imperative toplevel or mutually recursive nested closures in the future. A letrec for just a block of functions, like Fabio suggested, doesn't have this problem because you can put the block of mutually recursive functions between the imperative code just fine.

That said, I agree with Hisham that an implicit let rec for adjacent function declarations is unusual. If we do this I would prefer a more explicit syntax to be clear that the scoping rules are not the same as the Lua ones.

hishamhm:
Imperative toplevel makes sense for scripts, not for programs. It's unstructured programming.

hugomg:
Fair enough. For me the more appealing thing about imperative toplevels is that you simplify the language (same way to declare things inside and outside functions) and the implementation (global variables are just upvalues)

gligneul:
The imperative toplevel is appealing, but we shouldn't worry about it now since it is a huge change and it is not even in the roadmap.

spec/checker_spec.lua
```diff
@@ -108,7 +108,56 @@ describe("Titan type checker", function()
...
+            x3 = 11
+            x4 = 1.1
+            x5 = "11"
+            x6 = {}
+            x7 = {1, 2}
+            x8 = "a" .. x5
+            x9 = 1 + x3
+            x10 = not x2
+            x11: integer = 10.1
+        ]]
```
hugomg: Splitting these into separate assertions would be nice.

spec/checker_spec.lua
```diff
@@ -108,7 +108,56 @@ describe("Titan type checker", function()
...
+            function f(): integer return 10 end
+            x = {f()}
+        ]], [[
+            x = {10}
+            y = x[2]
+        ]]}
+        for _, c in ipairs(code) do
+            local ok, err = run_checker(c)
+            assert.falsy(ok)
+            assert.match("must be constant", err)
```
hugomg: I prefer using function calls instead of for loops because if you use `busted -v` the stack trace line numbers tell you which test was the one that failef

titan-compiler/checker.lua
```diff
@@ -689,43 +621,155 @@ local function checkfunc(node, st, errors)
...
+            const = const and isconst(field.exp)
+        end
+        return const
+
+    elseif tag == "Exp_Call" then
+        return false
+
+    elseif tag == "Exp_Var" then
+        if node.var._tag == "Var_Name" then
+            return true
```
hugomg: For now, I think we should consider these as not constants. The code generator currently evaluates the top-level in top down order so if we let it try to evaluate variables defined further down it is going to try to read an uninitialized variable.

We might be able to allow variables in a future version but in addition to changing the code generator we would also need to be careful about forbidding nonsensical mutual declarations such as 

    local a = b
    local b = a

hugomg (changes requested):

gligneul:
@hugomg, modifications done.

hugomg (approved):

hugomg:
Não sou muito fã de curry em linguagens dinâmicas. Lembro que quando tentei trabalhar com Javascript funcional o de vez em quando aparece um erro de "f is null" que era um saco de debugar porque não dava pra saber quem era pra ser o f.

hishamhm:
@hugomg Wat? :)

PR #100, opened 2017-12-18
Use semicolon as an optional field separator for records

gligneul:
Example:
```
record point
    x: float; y: float
end
```

### DIFF
--- a/spec/parser_spec.lua
+++ b/spec/parser_spec.lua
@@ -643,8 +643,8 @@ describe("Titan parser", function()
 
         assert_program_ast([[
             record List
-                p: {Point};
-                next: List;
+                p: {Point}
+                next: List
             end
         ]], {
             { _tag = "TopLevel_Record",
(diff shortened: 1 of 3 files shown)


spec/parser_spec.lua
```diff
@@ -643,8 +643,8 @@ describe("Titan parser", function()
 
         assert_program_ast([[
             record List
-                p: {Point}
-                next: List
+                p: {Point};
+                next: List;
```
hugomg: I think this deserves to be in its own test case instead of piggybacking on top of an existing test for recursive type declarations.

hugomg (approved):

spec/parser_spec.lua
gligneul: Agreed, I will add another test cases, please don't merge this yet.

hishamhm:
This should also be noted in the documentation (at the very least in the BNF section)

gligneul:
Done both changes, can someone verify and merge if everything is ok?

hishamhm (approved):

hishamhm:
Done! :)

PR #104, opened 2017-12-20
Multi-assignments in one statement (`local x,y = 1, 2`)

hishamhm:
This is the work-in-progress for supporting multi-assignments and multi-declarations.

While we don't have multiple returns, I'm requiring the declaration list and the expression list to have the same arity, no adjustments are performed. (Also, because we're always demanding an `= explist` at this point — I assume that in the future with nullable types a form `local v: nullable_type` without an expression should be allowed, and assign `v` to `nil`.)

Status:

* [x] multi-declarations in local variables
* [ ] multi-assignments (supporting `a, b = b, a`)
* [ ] multi-declarations in module variables (if they stay...)

Comments are welcome!


titan-compiler/checker.lua
```diff
@@ -241,17 +241,32 @@ function checkstat(node, st, errors)
...
-        if node.decl.type then
-          checkstat(node.decl, st, errors)
-          checkexp(node.exp, st, errors, node.decl._type)
-        else
-          checkexp(node.exp, st, errors)
-          node.decl._type = node.exp._type
-          checkstat(node.decl, st, errors)
+        for i = 1, #node.decl do
+            local decl_i = node.decl[i]
+            local exp_i = node.exp[i]
```
gligneul: Just a style comment: the `_i` suffix seems unnecessary.

titan-compiler/checker.lua
```diff
@@ -241,17 +241,32 @@ function checkstat(node, st, errors)
...
+                    checkexp(exp_i, st, errors, decl_i._type)
+                else
+                    checkexp(exp_i, st, errors)
+                    decl_i._type = exp_i._type
+                    checkstat(decl_i, st, errors)
+                end
+                exp_i = trycoerce(exp_i, decl_i._type, errors)
+                node.exp[i] = exp_i
+                checkmatch("declaration of local variable " .. decl_i.name,
+                    decl_i._type, exp_i._type, errors, decl_i._pos)
```
gligneul: Couldn't we move those 4 lines to the then case?
If the type come from the expression, we don't need to coerce and verify.

gligneul (commented):
This PR looks good to me, just minor comments.
The only thing I'm a bit worried is the confusion the syntax `local a, b: integer = ...` might cause. Looks like both variables are annotated as integers, but only `b` really is.

titan-compiler/checker.lua
```diff
@@ -241,17 +241,32 @@ function checkstat(node, st, errors)
...
-        if node.decl.type then
-          checkstat(node.decl, st, errors)
-          checkexp(node.exp, st, errors, node.decl._type)
-        else
-          checkexp(node.exp, st, errors)
-          node.decl._type = node.exp._type
-          checkstat(node.decl, st, errors)
+        for i = 1, #node.decls do
+            local decl = node.decls[i]
+            local exp = node.exps[i]
```
hishamhm, line 246: @gligneul I renamed the node attributes to `node.decls` and `node.exps` in plural, so I could rename the locals without causing confusion. (It's good practice to keep those array entries in the AST in plural anyway). Thanks for the nudge in the right direction!

hishamhm:
> The only thing I'm a bit worried is the confusion the syntax local a, b: integer = ... might cause. Looks like both variables are annotated as integers, but only b really is.

This is a very valid concern and indeed it's a sore point of the syntax. However, I think we'll have to live with it anyway if we're ever going to support multiple return values in the future.

hishamhm:
@gligneul Added a note in the manual about this point. I think it's the best we can do for now!

titan-compiler/checker.lua
gligneul: :v:

gligneul:
> @gligneul Added a note in the manual about this point. I think it's the best we can do for now!
@hishamhm yes indeed, I cannot think any other solution.

mascarenhas:
This got stale, and is also currently failing. I can fix it up if @hishamhm is currently busy, just let me know.

hishamhm:
@mascarenhas oh, I totally forgot about this. I think all of this is superseded by your current work, right? we can close this one then

mascarenhas:
> @mascarenhas oh, I totally forgot about this. I think all of this is superseded by your current work, right? we can close this one then

Yep, the `multiret` branch is a strict superset of this.

PR #105, opened 2017-12-20
Add TITAN_LIKELY and TITAN_UNLIKELY branch-prediction annotations

hishamhm:
Taking a page from the Linux kernel, introduce TITAN_LIKELY and
TITAN_UNLIKELY branch-prediction annotations. Here, we only add
the TITAN_LIKELY annotations to the branch that matches the
expected data type for type checks, and the TITAN_UNLIKELY branch
for runtime type errors.

These are enabled for GCC only, and are pass-through macros for
other compilers.

Anedoctal evidence: In the matmul1 benchmark adapted from Ravi,
this brought the execution time from 2.9s to 1.9s.


andremm (approved):
LGTM

PR #106, opened 2017-12-20
Disable unused function warning in MacOS

### LINKED ISSUE #87: Fix `-Wunused-function` warning when compiling with clang
When I compile a Titan program with clang, I get those warnings:

```
./foo.c:144:24: warning: unused function '_integer2str' [-Wunused-function]
inline static TString* _integer2str (lua_State *L, lua_Integer i) {
                       ^
./foo.c:150:24: warning: unused function '_float2str' [-Wunused-function]
inline static TString* _float2str (lua_State *L, lua_Number f) {
                       ^
```

These are related to the compiler internals and the end user shouldn't see them.

gligneul:
Fixes #87


titan-compiler/driver.lua
```diff
@@ -95,6 +95,15 @@ function driver.shared()
...
 
+function driver.cflags()
+    local uname = io.popen("uname"):read("*a")
+    if string.match(uname, "Darwin") then
+        return "-Wno-unused-function"
+    else
+        return ""
+    end
+end
+
```
andremm: Please fix the tests to also call `driver.cflags`.

andremm (dismissed):

titan-compiler/driver.lua
```diff
@@ -95,6 +95,15 @@ function driver.shared()
     return shared
 end
 
+function driver.cflags()
+    local uname = io.popen("uname"):read("*a")
```
hishamhm: This module is now doing the same `io.popen` in two places, and forking these subprocesses potentially multiple times, when multiple modules are compiled. We should cache the results, because in larger compilations this extra forking will add up. 

(Also, sorry to comment on an issue that existed prior to this patch, but this in an opportunity to fix it: both `io.popen` calls are not closing the fd and leaving up to the GC to deal with it... which my experience in LuaRocks shows it's not a good idea. Need to `io.popen` to a separate variable, read it then close it explicitly.)

gligneul: No problem, I will fix it.

andremm: Cool @hishamhm! When I wrote the `io.popen` in the first place I had a feeling that I should have split the operations, but still went for the easier way. Many thanks for sharing your LuaRocks experience that confirms the operations should be split.

mascarenhas (approved):

PR #108, opened 2017-12-26
Travis: cache ~/.luarocks

hishamhm:


hugomg:
I wasn't aware that this was even a thing but Travis gave it a green checkmark so it must be good.

hugomg (approved):

PR #110, opened 2017-12-27
Remove special cases from string coersions

gligneul:
Now concat will accept any value that can coerce to a string and numbers can be coersed to strings in any place.

The following code wasn't valid (now it should be):
```
function problem1(v: value)
    local s: string = v
    local s1: string = v .. v -- type error
    local s2: string = s .. s -- worked
end
function problem2(i: integer)
    local s1: string = i -- type error
    local s2: string = "" .. i -- worked
end
```

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -212,12 +212,12 @@ describe("Titan type checker", function()
             function fn()
                 local i: integer = 1
                 local s: string = "foo"
-                s = i
+                i = s
             end
         ]]
         local ok, err = run_checker(code)
         assert.falsy(ok)
-        assert.match("expected string but found integer", err)
+        assert.match("expected integer but found string", err)
     end)
 
     it("function can call another function", function()
@@ -236,12 +236,12 @@ describe("Titan type checker", function()
     it("catches mismatching types in arguments", function()
         local code = [[
             function fn(i: integer, s: string): integer
-                s = i
+                i = s
             end
         ]]
         local ok, err = run_checker(code)
         assert.falsy(ok)
-        assert.match("expected string but found integer", err)
+        assert.match("expected integer but found string", err)
     end)
 
     it("allows setting element of array as nil", function ()
@@ -330,14 +330,14 @@ describe("Titan type checker", function()
             function fn(x: integer): integer
                 local i: integer = 15
                 while i do
-                    local s: string = i
+                    local a: integer = "a"
                 end
                 return x
             end
         ]]
         local ok, err = run_checker(code)
         assert.falsy(ok)
-        assert.match("expected string but found integer", err)
+        assert.match("expected integer but found string", err)
     end)
 
     it("type-checks 'for' with a step", function()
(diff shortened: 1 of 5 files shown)


mascarenhas:
Not having automatic coercion from `value` to `string` on concatenation (but having it on assignment) was done on purpose, because the value might be something that has a `__concat` metamethod, and I would prefer if the semantics of `v .. v` would respect the metamethod in this case, and not just throw an error if `v` is not a string. I know there is `__tostring`, but it is not an "official" metamethod, so I was not feeling too guilty about ignoring it for the `local s: string = v` case.

For the `integer` to `string` case, I was keeping the coercion in concatenation just to follow Lua, and do not like to automatically coerce numbers to strings everywhere. Remember that anything that is valid for assignment is also valid for passing an argument to a function that expects a parameter of some different type.

gligneul:
> Not having automatic coercion from value to string on concatenation (but having it on assignment) was done on purpose, because the value might be something that has a __concat metamethod, and I would prefer if the semantics of v .. v would respect the metamethod in this case, and not just throw an error if v is not a string. I know there is __tostring, but it is not an "official" metamethod, so I was not feeling too guilty about ignoring it for the local s: string = v case.

I see, I can revert that for now.

> For the integer to string case, I was keeping the coercion in concatenation just to follow Lua, and do not like to automatically coerce numbers to strings everywhere. Remember that anything that is valid for assignment is also valid for passing an argument to a function that expects a parameter of some different type.

The special case where we coerce numbers to strings only in concatenation seems strange to me. We should either always require the explicit cast or always coerce to string.

mascarenhas:
> The special case where we coerce numbers to strings only in concatenation seems strange to me. We should either always require the explicit cast or always coerce to string.

Well, seems like Lua does the number->string coercion in other contexts where a string is expected. I think I would be happy either way.

hugomg:
I wouldn't mind having Titan be super strict about coercions and require explicit convercions between numbers and strings. That said, I also wouldn't mind requiring explicit conversions between int and float so take what I say with a grain of salt...

hishamhm:
Pragmatically speaking, number->string seems to me way less problematic than string->number, even more so on concatenation. (There's the idiom of coercing to string with `x..""` that's akin to coercing to boolean with `not not x` — one could say that the latter sort of establishes that operators can be polymorphic.)

hishamhm:
(I typed the above comment yesterday on the phone, before @hugomg's comment, but forgot to hit Send.)

Would this approach of making the operators polymorphic but the types themselves not auto-converting be a nice tradeoff of safety and convenience?

Non-exhaustive list:

```
integer * integer -> integer
float * integer -> float
integer * float -> float
integer / integer -> float
integer // integer -> integer
integer // float -> float (math.floor(x//y)+0.0 -- like Lua)
integer .. string -> string
string .. integer -> string
integer .. integer -> string (like Lua)
```

but these would fail without an explicit conversion:

```
my_integers[5] = 1 / 2
my_strings[1] = 10
```

PR #116, opened 2018-02-02
Some minor things

hugomg:
Some minor things that were clogging my todo list :)

@gligneul, can you confirm if it is OK to remove that return value from symtab:close_block?

### DIFF
--- a/titan-compiler/symtab.lua
+++ b/titan-compiler/symtab.lua
@@ -11,7 +11,6 @@ function symtab:open_block()
 end
 
 function symtab:close_block()
-    local last = self.blocks[#self.blocks]
     table.remove(self.blocks)
 end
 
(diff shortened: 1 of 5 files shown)


titan-compiler/symtab.lua
```diff
@@ -13,7 +13,6 @@ end
 function symtab:close_block()
     local last = self.blocks[#self.blocks]
     table.remove(self.blocks)
-    return last
```
gligneul: I think it is OK to remove this, but you should also remove the variable declaration at line 14.

gligneul (changes requested):
LGTM, only a minor detail.

gligneul (approved):

titan-compiler/symtab.lua
hugomg: Of course!

PR #119, opened 2018-02-18
Docs: always distinguish run-time and compile-time errors.

hishamhm:
I propose that whenever "error" is mentioned in the document, it should state explicitly whether it is run-time or compile-time (no matter how obvious it may be to us).


andremm (approved):
I totally agree with that!

PR #120, opened 2018-02-18
docs: always use the terms run-time and compile-time

andremm:
After PR #119 I propose to make a standard for the terms run-time and compile-time, that is, to always use run-time and compile-time in our docs instead of runtime, run time, compile-time, or compile time.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -93,9 +93,9 @@ run-time error). For example, if variable `i` has type `integer` and variable
 if `t` has type `{ value }` the assignment `t[1] = i` always succeeds.
 
 You can also use an expression that has type `value` in most contexts that
-expect something with another type, but this will generate a runtime check
+expect something with another type, but this will generate a run-time check
 that might fail. In our previous examples with `v`, `i`, and `t` the assignments
-`i = v` and `i = t[1]` are also allowed, but are checked at runtime to see
+`i = v` and `i = t[1]` are also allowed, but are checked at run-time to see
 if the value being assigned is really an integer (or a floating-point value
 that can be safely converted to an integer).
 
@@ -115,7 +115,7 @@ are not compatible:
         x: float
         y: float
     end
-    
+
     record PointV
         x: value
         y: value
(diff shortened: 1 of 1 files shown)


doc/manual.md
```diff
@@ -150,7 +150,7 @@ all dots to path separators, and then appending `.so`, for binary modules, or
 modules will correspond to `foo.{so|titan}`, `foo/bar.{so|titan}`, and `foo/bar/baz.{so|titan}`.
 The Titan compiler will recompile the module if its source is newer than its binary.
 
-Binary modules are looked up in the *runtime search path*, a semicolon-separated list
```
hugomg, line 153: Is this one supposed to be runtime or run-time?

hugomg (commented):
I agree what we should standardize things.

We can call the lifecycle phase "compile-time" and  "run-time" but should we still use "runtime" for the Titan libraries and execution environment?

doc/manual.md
andremm: run-time

Sorry, it looks like I forgot this one.

andremm:
> I agree what we should standardize things.

> We can call the lifecycle phase "compile-time" and "run-time" but should we still use "runtime" for the Titan libraries and execution environment?

I'm proposing to standardize things, that is, to use "compile-time" and "run-time" everywhere.

hugomg:
The more I think about this the more confused I get. I'll defer the decision to my esteemed colleagues.

hishamhm:
I think it is established to use run-time for the time when things run,
i.e. "things that happen at run-time" (as opposed to compile-time), and
runtime for the piece of software responsible for running things, i.e. "the
runtime is responsible for implementing run-time checks". I used this
convention in my thesis.

hishamhm:
See also:
https://en.m.wikipedia.org/wiki/Runtime_library
https://en.m.wikipedia.org/wiki/Run_time_(program_lifecycle_phase)

In my thesis I also used the hyphenated versions only as adjectives: e.g.
"compile-time errors are errors that happen at compile time".

Sorry if these criteria don't have the uniformity of using the same
spelling everywhere, but so is language. :) I think these various spellings
help clarifying the grammatical intent of each use (as long as they are
used consistently, of course!)

andremm:
Fair enough, I didn’t know about these differences. Thanks for the explanation!

Issue #121, opened 2018-02-18
Code generation crashes when generating code for casts


andremm:
I tried to compile
```
function f (a:integer, b:integer): boolean
    return a and b
end
```
and got the following stack traceback:
```
lua: ./titan-compiler/coder.lua:1105: attempt to concatenate a nil value (local 'cexp')
stack traceback:
	./titan-compiler/coder.lua:1105: in upvalue 'codeexp'
	./titan-compiler/coder.lua:657: in function <./titan-compiler/coder.lua:656>
	(...tail calls...)
	./titan-compiler/coder.lua:318: in function <./titan-compiler/coder.lua:314>
	(...tail calls...)
	./titan-compiler/coder.lua:1243: in upvalue 'codefuncdec'
	./titan-compiler/coder.lua:1656: in function 'titan-compiler.coder.generate'
	./titan-compiler/driver.lua:114: in function 'titan-compiler.driver.compile'
	./titan-compiler/driver.lua:107: in function 'titan-compiler.driver.compile_module'
	./titanc:46: in main chunk
	[C]: in ?
```

When I tried to change the return type from `boolean` to `integer`, as follows
```
function f (a:integer, b:integer): integer
    return a and b
end
```
then I got the following stack traceback:
```
lua: ./titan-compiler/util.lua:78: Internal compiler error: missing template variable CEXP
stack traceback:
	[C]: in function 'error'
	./titan-compiler/util.lua:78: in function <./titan-compiler/util.lua:75>
	[C]: in function 'string.gsub'
	./titan-compiler/util.lua:75: in function 'titan-compiler.util.render'
	(...tail calls...)
	./titan-compiler/coder.lua:318: in function <./titan-compiler/coder.lua:314>
	(...tail calls...)
	./titan-compiler/coder.lua:1243: in upvalue 'codefuncdec'
	./titan-compiler/coder.lua:1656: in function 'titan-compiler.coder.generate'
	./titan-compiler/driver.lua:114: in function 'titan-compiler.driver.compile'
	./titan-compiler/driver.lua:107: in function 'titan-compiler.driver.compile_module'
	./titanc:46: in main chunk
	[C]: in ?
```

andremm:
Fixed by PR #122.

PR #122, opened 2018-02-21
fix 'and' and 'or' between non-boolean values when used in a value co…

mascarenhas:
…ntext instead of control context (fixes issue #121)


andremm (approved):
LGTM

PR #125, opened 2018-02-25
Replace types.equals and types.Integer

gligneul:


hugomg:
I'm currently reviewing this and found in checker.lua:474 one instance where types.Integer was being used to mean BOTH actual integers and the invalid type but the tests didn't catch it due to invalid coverage. Please add the following test to the test suite:

    it("accepts correct use of length operator", function()
        local code = [[
            function fn(x: {integer}): integer
                return #x
            end
        ]]
        local ok, err = run_checker(code)
        assert.truthy(ok)
    end)

hugomg:
Neverming, actually I think the correct thing to to here is to always return Integer instead of Integer or Invalid, as I suggested previously.

I commited the needed fixes to the length-test-case branch. You can rebase from there. Other than that I think this is good to go.

hugomg (approved):

PR #131, opened 2018-03-01
Misc code cleanup from lablua branch.

hugomg:
I think it might be a good idea to merge these back to master, before we start making big changes to the other branch.

This PR is a bit larger than average but at least everything has been reviewed by both me and gligneul...

PR #132, opened 2018-03-02
Work towards removing coercions

hugomg:
Removed Value type to because it implied in adding tons of coercions everywhere (very complex). We should add this back later.

Restricted boolean operators (not, and,  or) to only work with booleans. This is part of the effort to remove overall polymorphism and also to remove the iscondition stuff. We might want to partially undo this later.

Removed many automatic coercions that made Titan not be  a subset of Lua. For example, coercing the return value to match the return type of a function. We should not add these back later.

----

One bit of the code that I need reviewed are [lines 106 and 107](https://github.com/titan-lang/titan/blob/64a8a118de4cf65e3d1d4beb70268daef527078d/titan-compiler/checker.lua#L706) in checker.lua. These used to be an `or` but I had to change them to an `and`. I think the `and` is correct but if it is then why were the tests passing beforehand? It is getting late here and I highly suspect that I am overlooking something in this part of the code.

----

(Btw, as I mentioned earlier, this PR is against the alblua2 branch to avoid messing up with PR #131)


mascarenhas:
> One bit of the code that I need reviewed are lines 106 and 107 in checker.lua. These used to be an or but I had to change them to an and. I think the and is correct but if it is then why were the tests passing beforehand? It is getting late here and I highly suspect that I am overlooking something in this part of the code.

The `and` also looks right to me, maybe tests were passing because explicit coercions were under-covered, and implicit coercions would not trigger the bug...

mascarenhas:
> Removed many automatic coercions that made Titan not be a subset of Lua. For example, coercing the return value to match the return type of a function. We should not add these back later.

If there is automatic promotion from, for example, integers to floats it is odd if you cannot return an integer from a function that specifies that it must return a float. I am also not convinced that `being a subset of Lua` is a useful design constraint!

mascarenhas:
> Restricted boolean operators (not, and, or) to only work with booleans. This is part of the effort to remove overall polymorphism and also to remove the iscondition stuff. We might want to partially undo this later.

With option types this removes a very useful idiom: `foo or <exp>` to initialize optional parameters (right now this idiom already works with arrays, with `arr[i] or <exp>` being a way to take a value out of an array or a default value if it is `nil`. You also remove the "ternary operator".

hugomg:
> If there is automatic promotion from, for example, integers to floats it is odd if you cannot return an integer from a function that specifies that it must return a float.

I think these are different "kinds" of coercions. With arithmetic operators the coercions are an implementation detail we use for implementing operator overloading. We could also imagine `+` as being overloaded between `(integer, integer) -> integer`, `(float, integer) -> float` and so on, with no implicit coercions.

On the other hand, when Titan coerces the return value of a function or the value being assigned to a function or array, then Titan is introducing extra coercions that are not present in the equivalent Lua code.

> I am also not convinced that being a subset of Lua is a useful design constraint!

I agree that this is a valid point of discussion. But I still think it is worth getting rid of this code in this branch at least for now because these coercions are adding extra complexity that isn't strictly necessary for the code generation experiments we want to run in the very near future.

> With option types this removes a very useful idiom [...] to initialize optional parameters. [...] You also remove the "ternary operator".

I am aware of that and think that these idioms are OK even in the more restrictive view of Titan as a strict subset of Lua. However, I think that for now this branch can live without these features and doing so removes some of the trickiest code in  the type checker and code generation. Right now my biggest priority is to pare down this side branch of Titan to the absolute most barebones version needed to run some experiments on.

PR #133, opened 2018-03-03
For loop scope

hugomg:


gligneul (approved):

hishamhm:
This can also be ported into master, right? (Sorry ifit already has and I missed it!)

hugomg:
No, you made a great question!

I think this particular change could (and probably should) also be added to master but I just couldn't think of how to do that, from a git point of view... Should we create a different PR to cherry-pick some of the changes from the lablua branch onto master? Or something else? (We can move this discussion to slack if you prefer)

hishamhm:
yes, a PR with cherry-picked commits would work fine (and hopefully not too much work!)

PR #138, opened 2018-03-04
Refactor make visitor

hugomg:
- Get rid of make_visitor
- Standardize default behavior as `error("impossible")
- Fix a harmless bug in check_stat


hishamhm:
Why the backpedalling on this? Is it to make it easier to bootstrap the compiler?

hugomg:
It turns out that in its current form, the "visitor" was just a fancy switch statement. The only thing it effectively did was put that `error("impossible")` in the end, which we found to be not worth the complexity.

In the future we might create a real Ast_visitor that does a full traversal of the AST but we think it would be better to have some more passes of the compiler written by hand before committing to that abstraction.

gligneul (dismissed):

gligneul (approved):

PR #141, opened 2018-03-06
Lpeglabel15

sqmedeiros:
Merge this branch into master

### DIFF
--- a/titan-compiler/lexer.lua
+++ b/titan-compiler/lexer.lua
@@ -22,7 +22,10 @@ local C, Cb, Cg, Ct, Cmt, Cc = lpeg.C, lpeg.Cb, lpeg.Cg, lpeg.Ct, lpeg.Cmt, lpeg
 local T = lpeg.T
 
 local syntax_errors = require "titan-compiler.syntax_errors"
-local labels = syntax_errors.label_to_int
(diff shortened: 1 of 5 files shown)


hugomg (approved):
Looks good to me

titan-compiler/lexer.lua
```diff
@@ -22,7 +22,10 @@ local C, Cb, Cg, Ct, Cmt, Cc = lpeg.C, lpeg.Cb, lpeg.Cg, lpeg.Ct, lpeg.Cmt, lpeg
 local T = lpeg.T
 
 local syntax_errors = require "titan-compiler.syntax_errors"
-local labels = syntax_errors.label_to_int
+local labels = {}
```
gligneul, line 25: We don't need this table anymore. Couldn't we just remove this piece of code and use the actual label now?
`s/labels.\([A-Za-z_]*\)/"\1"/g`

gligneul (dismissed):
Looks good, just a minor detail I think we should fix now.

titan-compiler/lexer.lua
hugomg: I can fix this in a subsequent pull request. If this is the only problem I think we can merge it now.

PR #142, opened 2018-03-06
Scope analysis

hugomg:
Separate the lexical scope analysis into a separate pass. I think the way we were treating check_decl in the type checker was fundamentally bug prone...

(work in progress -- see current commit messages)


titan-compiler/scope_analysis.lua
```diff
@@ -0,0 +1,343 @@
...
+        bind_names_exp(var.exp, st, errors)
+
+    else
+        error("impossible")
+    end
+end
+
+bind_names_exp = function(exp, st, errors)
+    local tag = exp._tag
+    if     tag == ast.Exp.Nil then
```
gligneul, line 277: We could merge those "nothing to do" like `bind_names_type`.

gligneul (approved):

PR #169, opened 2018-03-20
Lua5.4

hugomg:
Updating to Lua 5.4 will let us take advantage of new features. Multiple uservalues per userdata should be specially useful.

The changelog for this is a bit hard to go through because it includes all the changes between Lua 5.3 and Lua 5.4 so it might be easier to look at the individual commits, skipping the first one.

Overall it was a very smooth process. The only Lua patch needed was to redefine LUAI_FUNC in luaconf (I got rid of the "titan path" patches related to modules). And in the code generation it seemed that the only thing affected was that the Lua stack is now an array of StackValue and that the meaning of the ttype macro changed.

However, additional issues might pop up when we manage to fix the automated tests, which broke again...


hugomg:
No additional problems showed up now that the full test suite is running again. I'm optimistic for now...

gligneul (approved):

PR #172, opened 2018-03-22
multiple return values from functions, plus multiple assignment

mascarenhas:
Right now just a stub, with a new `Ast.ExpAdjust` node for parenthesized function calls, plus changing `Ast.StatReturn` to take an expression list. The typechecker complains if the number of expressions in this expression list is different from the number of return types, not taking into account the last expression returning multiple values right now. The code generator just generates code for the first return expresssion, enough to make our current tests pass.

Roadmap: 

* [x] fix typechecker to take into account trailing calls in expression lists, both in `StatReturn` and in the argument lists of `ExpCall`
* [x] make code generator generate proper signatures for functions with multiple return values, and make `StatReturn` work with them
* [x] add multiple local variable declaration, and make it interact with multiple return values; multiple declaration is lenient with extra returned values
* [x] fix code generation for argument lists to take multiple return values in the trailing position and pass them along; maintain strict arity check
* [ ] add multiple assignment, and make it interact with multiple return values; multiple assignment is lenient with extra returned values
* [x] use multiple return values in array constructors' expression lists

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -1475,6 +1475,19 @@ describe("Titan type checker", function()
         assert_type_check([[
             function f()
                 local x: float = 10, 20, "foo"
+                x = 10, 20, "foo"
+            end
+        ]])
+    end)
+
+    it("typechecks multiple declarations and assignment", function()
+        assert_type_check([[
+            function g(): (integer, string)
+                return 20, "foo"
+            end
+            function f()
+                local x: float, y: integer, z: string = 10, g()
+                x = 10, g()
             end
         ]])
     end)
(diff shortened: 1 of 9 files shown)


hishamhm:
Yay!

Made your roadmap into convenient checkboxes :)

mascarenhas:
The only thing missing right now is proper multiple assignment, right now we only have multiple declaration of local variables. Multiple local variable declaration is easier because the expressions in the right hand side cannot use the variables on the left hand side, so we can generate the code for each lvalue/rvalue par sequentially. For multiple assignment we first need to store the rvalues in temporaries, then do the assignments.

mascarenhas:
I removed the "wip" label because what is committed right now is good enough for merging to master, IMHO.

hishamhm (approved):
Awesome -- and yes, let's try to break apart those major pieces and work and try to start merging in stuff more gradually. I need to do the same for the `foreign` branch!

PR #174, opened 2018-03-26
Store a 'vararg' bit in functions

hishamhm:
Not yet used by Titan functions, but already useful for (and used by) FFI functions.

Cherry-picked from the `foreign` branch so we can start pushing isolated changes to `master`.


hugomg:
What is the meaning of the vararg bit? That there are additional vararg parameters after the `params` list? Or does the presence of the varag list invalidate the `params` field?

hishamhm:
Additional vararg arguments (ie, `...` after regular arguments), of course -- why would we invalidate the params list? :)

mascarenhas:
I think @hugomg is asking about a `...` parameter vs the `vararg` bit, and I am sure @hishamhm proposal is to *not* have a `...` parameter in the AST, even if we reintroduce this syntax for vararg functions. A function with a parameter list containing three positional parameters plus `...` would have the three parameters in the `params` list and the `vararg` bit set, instead of four parameters, one of them being `...`.

mascarenhas (approved):

PR #176, opened 2018-03-26
Implement functions with zero return values

hugomg:
I think functions that return nothing are better "void" functions than functions that return "nil".

This also changes the test for "control reaches end of non-void" function to require a "return nil" for functions that return nil, and only make the return statement be optional in functions that return 0 values.

### DIFF
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -738,12 +738,11 @@ describe("Titan type checker", function()
             end
 
             local function g(): integer
-                local x = 1 + f()
-                return x
+                return 1 + f()
             end
         ]])
         assert.falsy(prog)
-        assert.match("expected integer but found invalid type", errs)
+        assert.match("void instead of a number", errs)
     end)
 
     it("detects attempts to call non-functions", function()
--- a/spec/checker_spec.lua
+++ b/spec/checker_spec.lua
@@ -642,8 +642,48 @@ describe("Titan type checker", function()
         assert.match("'for' step expression", errs)
     end)
 
-    it("detects when a functionr returns the wrong type", function()
-        local code, errors = run_checker([[
+    it("detects wrong number of return values", function()
+        local prog, errs = run_checker([[
+            function f(): ()
+                return 1
+            end
+        ]])
+        assert.falsy(prog)
+        assert.match(
+            "returning 1 value(s) but function expects 0", errs,
+            nil, true)
+
+        local prog, errs = run_checker([[
+            function f(): integer
+                return
+            end
+        ]])
+        assert.falsy(prog)
+        assert.match(
+            "returning 0 value(s) but function expects 1", errs,
+            nil, true)
+    end)
+
+    it("accepts functions that return 0 values", function()
+        local prog, errors = run_checker([[
+            function f(): ()
+                return
+            end
+        ]])
+        assert.truthy(prog)
+    end)
+
+    it("accepts functions that return 1 value", function()
+        local prog, errors = run_checker([[
+            function f(): integer
+                return 17
+            end
+        ]])
+        assert.truthy(prog)
+    end)
+
+    it("detects when a function returns the wrong type", function()
+        local prog, errors = run_checker([[
             function fn(): integer
                 return "hello"
             end
@@ -652,7 +692,7 @@ describe("Titan type checker", function()
         assert.match("types in return statement do not match, expected integer but found string", errors)
     end)
 
-    it("detects nil returns on non-nil functions", function()
+    it("detects missing return statements", function()
         local code = {[[
             function fn(): integer
             end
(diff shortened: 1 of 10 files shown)


spec/checker_spec.lua
```diff
@@ -688,10 +728,24 @@ describe("Titan type checker", function()
...
+            local function f(): ()
+            end
+
+            local function g(): integer
+                local x = 1 + f()
+                return x
+            end
+        ]])
+        assert.falsy(prog)
+        assert.match("expected integer but found invalid type", errs)
```
gligneul: Maybe we should add an `void` type? I think this message would be clearer.

titan-compiler/checker.lua
```diff
@@ -112,7 +112,7 @@ check_type = function(typ, errors)
         return types.T.Array(check_type(typ.subtype, errors))
 
     elseif tag == ast.Type.Function then
-        if #typ.argtypes ~= 1 then
+        if #typ.rettypes >= 2 then
             error("functions with 0 or 2+ return values are not yet implemented")
```
gligneul: We should update the assert message to `"function with 2+ return...`

titan-compiler/checker.lua
```diff
@@ -147,7 +147,7 @@ check_toplevel = function(tl_node, errors)
         end
 
     elseif tag == ast.Toplevel.Func then
-        if #tl_node.rettypes ~= 1 then
+        if #tl_node.rettypes >= 2 then
             error("functions with 0 or 2+ return values are not yet implemented")
```
gligneul: Same here.

titan-compiler/checker.lua
```diff
@@ -644,8 +653,12 @@ check_exp = function(exp, errors, typehint)
                     "function %s called with %d arguments but expects %d",
                     fname, nargs, nparams)
             end
-            assert(#ftype.rettypes == 1)
-            exp._type = ftype.rettypes[1]
+            assert(#ftype.rettypes <= 1)
+            if #ftype.rettypes >= 1 then
+                exp._type = ftype.rettypes[1]
+            else
+                exp._type = types.T.Invalid()
```
gligneul: types.T.Void()?

titan-compiler/coder.lua
```diff
@@ -337,23 +342,41 @@ generate_program = function(prog, modname)
...
+                    set_ret = util.render([[
+                        ${SET_SLOT}
+                        api_incr_top(L);
+                    ]], {
+                        SET_SLOT = set_slot(ret_typ, "s2v(L->top)", "ret")
+                    })
+                else
+                    error("not implemented")
+                end
+
```
gligneul, line 361: s/\n//

titan-compiler/coder.lua
```diff
@@ -337,23 +342,41 @@ generate_program = function(prog, modname)
                     table.insert(args, arg_name)
                 end
 
+                local ret_init, set_ret
+                if #tl_node._type.rettypes == 0 then
+                    ret_init = ""
+                    set_ret = ""
+                elseif #tl_node._type.rettypes == 1 then
+                    local ret_typ = tl_node._type.rettypes[1]
+                        ret_init = c_declaration(ctype(ret_typ), "ret") .. " ="
```
gligneul: There is an extra indentation level here.

titan-compiler/coder.lua
```diff
@@ -627,16 +650,26 @@ generate_stat = function(stat)
...
+            ${IGNORE_RESULT}
         ]], {
             STATS = cstats,
-            VALUE = cvalue
+            IGNORE_RESULT = ignore_result
         })
 
     elseif tag == ast.Stat.Return then
-        local cstats, cvalue = generate_exp(stat.exp)
+        assert(#stat.exps == 1)
```
gligneul: What happens if we have a return with no exps (eg. void function)?

titan-compiler/coder.lua
```diff
@@ -804,20 +837,27 @@ generate_exp = function(exp) -- TODO
...
                 local cstats = util.render([[
                     ${ARG_STATS}
-                    ${TMP_DECL} = ${FUN_NAME}(${ARGS});
+                    ${TMP_INIT} ${FUN_NAME}(${ARGS});
                 ]], {
                     FUN_NAME  = tl_node._titan_entry_point,
                     ARG_STATS = table.concat(arg_cstatss, "\n"),
                     ARGS      = table.concat(arg_cvalues, ", "),
-                    TMP_DECL = tmp_decl,
+                    TMP_INIT = tmp_init,
```
gligneul: missing a space before `=`

gligneul (changes requested):
Just minor stuff and some questions

gligneul (approved):

PR #178, opened 2018-03-29
Code generation for multiple assignment

mascarenhas:
Cherry-picked from my broken `multiret` branch.

* [x] Make arity checks in assignment and return strict
* [x] Refactor common parts of `codesingleassignment` and `codemultipleassignment`; they should be kept separate to avoid going through an extra temporary in single assignment


mascarenhas:
This should be good to go.

hishamhm (approved):

PR #180, opened 2018-03-29
Implement table.insert and table.remove

hugomg:


titan-compiler/ast.lua
```diff
@@ -23,6 +23,7 @@ declare_type("Toplevel", {
     Var    = {"loc", "decl", "value"},
     Record = {"loc", "name", "field_decls"},
     Import = {"loc", "localname", "modname"},
+    Builtin = {"loc", "name"},
```
gligneul: Could we fix the alignment in the lines on top?

titan-compiler/checker.lua
```diff
@@ -82,6 +98,11 @@ end
 --
 
 check_program = function(prog, errors)
+    -- Ugh
```
gligneul: This is ugly indeed, but I think we should add a more meaningful comment here.
Since we don't plan to fix this soon, I think it would be nice to document a better solution here (eg. having an algebraic type for declarations).

titan-compiler/checker.lua
```diff
@@ -367,6 +404,32 @@ check_var = function(var, errors)
     end
 end
 
+
```
gligneul: s/\n//

titan-compiler/checker.lua
```diff
@@ -367,6 +404,32 @@ check_var = function(var, errors)
...
+local function check_exp_callfunc_builtin(exp, errors)
```
gligneul: Could we move this function to the top of the file?
I think it is strange to have it between others `check_` functions (despite it is only used by call_exp).

titan-compiler/checker.lua
```diff
@@ -367,6 +404,32 @@ check_var = function(var, errors)
...
+    local fexp = exp.exp
+    local args = exp.args
+    local builtin_name = fexp._type.builtin_decl.name
+    if builtin_name == "table.insert" then
+        check_arity("table.insert arguments", 2, #args, errors, exp.loc)
+        check_exp(args[1], errors, nil)
+        check_is_array("table.insert first argument",
```
gligneul, line 423: I think we need a little bit of meta-programming generalization here.
It won't be a blocker for this PR, but it will be necessary as others builtin functions appears.

titan-compiler/coder.lua
```diff
@@ -761,6 +761,85 @@ generate_var = function(var)
...
+        ${CSTATS_T}
+        ${CSTATS_V}
+        ${UI_DECL} = luaH_getn(${CVALUE_T});
+        ${SIZE_DECL} = ${CVALUE_T}->sizearray;
+        if (${SIZE} <= ${UI}) {
+            if (${SIZE} < 8) {
+                /* avoid infinite loop if sizearray == 0 */
+                ${SIZE} = 8;
+            }
+            while (${SIZE} <= ${UI}) {
```
gligneul, line 785: There is no other way to compute this without a loop?

titan-compiler/coder.lua
```diff
@@ -912,7 +1001,17 @@ generate_exp = function(exp) -- TODO
 
         local op = exp.op
         if op == "#" then
-            error("not implemented yet")
+            local tmp = tmp_name()
```
gligneul: What happens if we use the length operator with a string?

titan-compiler/checker.lua
hugomg: I would prefer keeping these closer to check_exp, since that is the only place where they will be used.

titan-compiler/coder.lua
hugomg: I think this loop should only ever need to go through one iteration but am not 100% sure about that so I played safe and went with the while loop

titan-compiler/coder.lua
hugomg: Something horrible would happen. Good catch!

titan-compiler/checker.lua
hugomg: In this specific instance the problem would also go away if we had parametric polymorphism. I think we should see what happens with additional builtins before biting the bullet here.

titan-compiler/builtins.lua
```diff
@@ -0,0 +1,8 @@
+local ast = require "titan-compiler.ast"
+
+local builtins = {}
```
gligneul: Could we format like that?
```
local builtins = {
    ["table_insert"] = ast.Toplevel.Builtin(false, "table.insert"),
    ["table_remove"] = ast.Toplevel.Builtin(false, "table.remove"),
}

titan-compiler/checker.lua
gligneul: Agreed

titan-compiler/coder.lua
gligneul: Couldn't this while be replaced with an else? If we increment the array by one, doubling the size should always work.

gligneul (approved):

PR #181, opened 2018-03-29
Some fixes to `titanc` and examples

mascarenhas:
The `driver.compile` function was not taking into account dots in the module name, so `titanc` was generating files with wrong names. Also renamed `titanc` to `titan`, to reflect the change in the help text for it.

### DIFF
--- a/examples/test_fill_table.lua
+++ b/examples/test_fill_table.lua
@@ -1,4 +1,4 @@
-local mod = require "artisanal"
+local mod = require "examples.artisanal"
 
 local f = mod.filltable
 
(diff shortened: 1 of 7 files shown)


mascarenhas:
BTW, I ended up making this dependent on PR #178, but as this is not critical it is ok to only merge this after that PR commits are in master.

hugomg:
I would prefer to continue calling the compiler titanc. Iirc, we decided to name it like that to be more consistent with other programming languages

hishamhm:
also, eventually I'd like to have a `titan` catch-all command that drives `titanc` and does more things, like `go` (`titan fmt`, `titan get`, etc.)

hishamhm (approved):

hishamhm:
I went ahead and merged this without the `titanc` -> `titan` rename (learning to be more "agile" :laughing: ), this is a fix worth having now (was already biting me in `foreign`) -- we can revisit the executable name issue later!

PR #187, opened 2018-04-05
Context/Cvar cleanup

hugomg:
Next step is actually implementing the GC :)

### DIFF
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -40,6 +40,20 @@ describe("Titan coder", function()
         ]])
     end)
 
+    it("Verify the argument numbers", function()
+        run_coder([[
+            function f(x: integer): integer
+                return x
+            end
+        ]], [[
+            local ok, err = pcall(test.f)
+            assert(string.find(err,
+                "wrong number of arguments to function, " ..
+                "expected 1 but received 0",
+                nil, true))
+        ]])
+    end)
+
     it("Verify the argument's tag", function()
         run_coder([[
             function f(x: float): float
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -41,15 +41,25 @@ describe("Titan coder", function()
     end)
 
     it("Verify the argument numbers", function()
-        run_coder([[
+        local code = [[
             function f(x: integer): integer
                 return x
             end
-        ]], [[
+        ]]
+
+        run_coder(code, [[
             local ok, err = pcall(test.f)
             assert(string.find(err,
                 "wrong number of arguments to function, " ..
                 "expected 1 but received 0",
+                nil, true))
+        ]])
+
+        run_coder(code, [[
+            local ok, err = pcall(test.f, 10, 20)
+            assert(string.find(err,
+                "wrong number of arguments to function, " ..
+                "expected 1 but received 2",
                 nil, true))
         ]])
     end)
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -336,12 +336,7 @@ local function generate_titan_entry_point(tl_node)
             ARR_DECL = c_declaration(ctx.upvalues.array),
         }))
     end
-    assert(tl_node.block._tag == ast.Stat.Block)
-    ctx:begin_scope()
-    for _, stat in ipairs(tl_node.block.stats) do
-        table.insert(body, generate_stat(stat, ctx))
-    end
-    ctx:end_scope()
+    table.insert(body, generate_stat(tl_node.block, ctx))
 
     return util.render([[
         static ${RET} ${NAME}(
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -357,7 +357,7 @@ local function generate_lua_entry_point(tl_node)
     local ctx = Context.new()
 
 
-    local base = ctx:new_cvar("StackValue*")
+    local base = ctx:new_cvar("StackValue*", "ci->func")
     local set_base = util.render("${BASE_DECL} = L->ci->func;", {
         BASE_DECL = c_declaration(base)
     })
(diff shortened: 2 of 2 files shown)


spec/coder_spec.lua
```diff
@@ -40,6 +40,20 @@ describe("Titan coder", function()
         ]])
     end)
 
+    it("Verify the argument numbers", function()
```
gligneul, line 43: Do we verify if we pass more arguments than necessary (eg. `test.f(1, 2)`)?
I think this is a relevant test, even if it isn't an error.

titan-compiler/coder.lua
```diff
@@ -271,273 +300,385 @@ local function set_heap_slot(typ, dst, src, parent)
...
-                    error("not implemented")
-                end
 
-                table.insert(function_definitions,
+    local base = ctx:new_cvar("StackValue*", "ci->func")
+    local set_base = util.render("${BASE_DECL} = L->ci->func;", {
+        BASE_DECL = c_declaration(base)
+    })
+
+    local function argslot(i)
```
gligneul: We could move this function near its use (after nargs part).

titan-compiler/coder.lua
```diff
@@ -271,273 +300,385 @@ local function set_heap_slot(typ, dst, src, parent)
...
+    local arg_vars  = {}
+    local get_args = {}
+    for i, param in ipairs(tl_node.params) do
+        local arg = ctx:new_tvar(param._type)
+        table.insert(arg_vars, arg)
+        table.insert(get_args, util.render([[
+            ${ARG_DECL} = ${SLOT_VALUE};
+        ]], {
+            ARG_DECL = c_declaration(arg),
+            SLOT_VALUE = get_slot(param._type, argslot(i)),
```
gligneul, line 422: We are calling argslot here again and replicating the `s2v(x1 + 1)` code.
Shouldn't we use the `slot.name` from previous loop? Is gcc able to optimize this?

titan-compiler/coder.lua
```diff
@@ -271,273 +300,385 @@ local function set_heap_slot(typ, dst, src, parent)
...
+        SET_BASE = set_base,
+        CHECK_NARGS = check_nargs,
+        CHECK_TYPES = table.concat(check_types, "\n"),
+        GET_ARGS = table.concat(get_args, "\n"),
+        SET_RETURN = set_return,
+        NRET = c_integer(#tl_node._type.rettypes),
+    })
+end
+
+local function generate_luaopen_upvalue_array(prog, ctx)
```
gligneul: I think we should rename upvalue to something else since upvalue is a Lua implementation detail (the way Lua implements environments).
Maybe call it modvars?
If we don't think about a better name we should at least add a comment.

titan-compiler/coder.lua
```diff
@@ -271,273 +300,385 @@ local function set_heap_slot(typ, dst, src, parent)
...
-                        TITAN_ENTRY_POINT = tl_node._titan_entry_point,
-                        RET_INIT = ret_init,
-                        ARGS_DECL = table.concat(args_decl, "\n"),
-                        ARGS = table.concat(args, ", "),
-                        SET_RET = set_ret,
-                        NRET = c_integer(#tl_node._type.rettypes)
+                        UPVALUES = ctx.upvalues.table.name,
+                        FUNC = func.name,
+                        FUNC_DECL = c_declaration(func),
+                        SET_SLOT = set_heap_slot( -- TODO
```
gligneul, line 508: TODO what?

titan-compiler/coder.lua
```diff
@@ -271,273 +300,385 @@ local function set_heap_slot(typ, dst, src, parent)
...
-                [[Table *titan_globals = hvalue(&clCvalue(s2v(L->ci->func))->upvalue[0]);]])
-        end
+    return table.concat(parts, "\n")
+end
 
-        for _, tl_node in ipairs(prog) do
-            if tl_node._global_index then
-                local arr_slot = util.render([[ &titan_globals->array[${I}] ]], {
+local function generate_luaopen_exports_table(prog, ctx)
+
```
gligneul: s/\n//

titan-compiler/coder.lua
```diff
@@ -271,273 +300,385 @@ local function set_heap_slot(typ, dst, src, parent)
...
+    ]], {
+        N = c_integer(n_exported_functions),
+        PARTS = table.concat(parts, "\n")
+    })
+end
+
+local function generate_luaopen(prog, modname)
+    local ctx = Context.new()
+    -- TODO: this is actually an array of Value
+    local table_typ = types.T.Array( types.T.Integer() )
```
gligneul: `s/types.T.Array( types.T.Integer() )/types.T.Array(types.T.Integer())/`

titan-compiler/coder.lua
```diff
@@ -271,273 +300,385 @@ local function set_heap_slot(typ, dst, src, parent)
...
+            return 1;
+        }
+    ]], {
+        MODNAME = modname,
+        ALLOCATE_UPVALUES = allocate_upvalues,
+        INIT_UPVALUES = init_upvalues,
+        INIT_EXPORTS = init_exports,
+    })
+end
+
```
gligneul: s/\n//

gligneul (changes requested):

titan-compiler/coder.lua
hugomg: I am not worried about this very much because arithmetic is very cheap.

titan-compiler/coder.lua
hugomg: I honestly don't remember... But this TODO is gone in the current versioni of the code. I must have fixed whatever it was.

spec/coder_spec.lua
hugomg: I added an extra test for that, just in case.

gligneul (approved):

PR #195, opened 2018-04-10
Sum benchmark

gligneul:


benchmarks/sum/lua.lua
```diff
@@ -0,0 +1,11 @@
+local _M = {}
+function _M.sum(arr, n)
+    local s = 0.0
+    for i = 1, n do
```
hugomg: Shadowing the "i" is a bit confusing. What about `for rep = 1, nrep`?

benchmarks/sum/titan_no_checks.c
```diff
@@ -0,0 +1,127 @@
+/* This file was generated by the Titan compiler. Do not edit by hand */
```
hugomg: The titan compiler is still evolving very fast. I think it might be a good idea to avoid committing these "artisanal" .c files for now.

hugomg (changes requested):

benchmarks/sum/lua.lua
gligneul: Done

benchmarks/sum/titan_no_checks.c
gligneul: Agreed

hugomg (approved):

PR #196, opened 2018-04-10
Make the generated code less cluttered

hugomg:
Some miscelaneous changes aimed at making the generated code more compact and easier to read.

The biggest change is that the error paths are encapsulated inside functions now. Another notable change is that runtime error messages now talk only about Lua tags instead of Titan types (it is clever enough to show integer/float instead of number though).

One thing that probably needs to be improved is hot to handle the warning-supression flags. Right now they are split between the CFLAGS in c_compiler.lua and \#pragma declarations in the generated C code. Ideally we should only do it one way but which way is better? Do those \#pragma declarations also apply to non-clang compilers?

### DIFF
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -78,6 +78,21 @@ describe("Titan coder", function()
         ]])
     end)
 
+    it("Verify the argument's tag (float/integer vs number)", function()
+        run_coder([[
+            function f(x: float): float
+                return x
+            end
+        ]], [[
+            local ok, err = pcall(test.f, 10)
+            assert(string.find(err,
+                "wrong type for argument x at line 1, " ..
+                "expected float but found integer",
+                nil, true))
+        ]])
+    end)
+
+
     describe("Expressions:", function()
 
         it("Constants", function()
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -90,8 +90,19 @@ describe("Titan coder", function()
                 "expected float but found integer",
                 nil, true))
         ]])
-    end)
 
+        run_coder([[
+            function f(x: integer): integer
+                return x
+            end
+        ]], [[
+            local ok, err = pcall(test.f, 3.14)
+            assert(string.find(err,
+                "wrong type for argument x at line 1, " ..
+                "expected integer but found float",
+                nil, true))
+        ]])
+    end)
 
     describe("Expressions:", function()
 
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -796,11 +796,14 @@ describe("Titan coder", function()
             ]])
         end)
 
-        it("can set wrongly typed arrays in set", function()
+        it("checks type tags in set", function()
             run_coder(array_get_set, [[
                 local arr = {10, 20, "hello"}
-                test.set(arr, 3, 123)
-                assert(123 == arr[3])
+
+                local ok, err = pcall(test.set, arr, 3, 20)
+                assert(not ok)
+                assert(
+                    string.find(err, "wrong type for array element", nil, true))
             ]])
         end)
 
--- a/titan-compiler/c_compiler.lua
+++ b/titan-compiler/c_compiler.lua
@@ -5,7 +5,7 @@ local c_compiler = {}
 
 c_compiler.LUA_SOURCE_PATH = "./lua/src"
 c_compiler.CFLAGS_BASE = "--std=c99 -g "
-c_compiler.CFLAGS_WARN = "-Wall -Wno-unused-function -Wno-parenthesis-equality"
+c_compiler.CFLAGS_WARN = "-Wall -Wno-unused-function -Wno-parentheses-equality"
 c_compiler.CFLAGS_OPT = "-O2"
 c_compiler.CC = "cc"
 
--- a/titan-compiler/c_compiler.lua
+++ b/titan-compiler/c_compiler.lua
@@ -4,7 +4,7 @@ local util = require "titan-compiler.util"
 local c_compiler = {}
 
 c_compiler.LUA_SOURCE_PATH = "./lua/src"
-c_compiler.CFLAGS_BASE = "--std=c99 -Wall -g"
+c_compiler.CFLAGS_BASE = "--std=c99 -Wall -g -Wno-unused-function"
 c_compiler.CFLAGS_OPT = "-O2"
 c_compiler.CC = "cc"
 
--- a/titan-compiler/c_compiler.lua
+++ b/titan-compiler/c_compiler.lua
@@ -4,7 +4,8 @@ local util = require "titan-compiler.util"
 local c_compiler = {}
 
 c_compiler.LUA_SOURCE_PATH = "./lua/src"
-c_compiler.CFLAGS_BASE = "--std=c99 -Wall -g -Wno-unused-function"
+c_compiler.CFLAGS_BASE = "--std=c99 -g "
+c_compiler.CFLAGS_WARN = "-Wall -Wno-unused-function -Wno-parenthesis-equality"
 c_compiler.CFLAGS_OPT = "-O2"
 c_compiler.CC = "cc"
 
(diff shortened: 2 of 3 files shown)


gligneul:
> Ideally we should only do it one way but which way is better? 

I think we should centralize it in the flags in order to make the output less bloated.

> Do those #pragma declarations also apply to non-clang compilers?

Each compiler have its own set of #pragma flags.

spec/coder_spec.lua
```diff
@@ -78,6 +78,21 @@ describe("Titan coder", function()
...
+        ]], [[
+            local ok, err = pcall(test.f, 10)
+            assert(string.find(err,
+                "wrong type for argument x at line 1, " ..
+                "expected float but found integer",
+                nil, true))
+        ]])
+    end)
+
+
```
gligneul: s/\n//

spec/coder_spec.lua
```diff
@@ -78,6 +78,21 @@ describe("Titan coder", function()
         ]])
     end)
 
+    it("Verify the argument's tag (float/integer vs number)", function()
+        run_coder([[
+            function f(x: float): float
```
gligneul, line 83: Shoudn't we also have the opposite test?
Expected int but found float?

titan-compiler/c_compiler.lua
```diff
@@ -4,7 +4,7 @@ local util = require "titan-compiler.util"
 local c_compiler = {}
 
 c_compiler.LUA_SOURCE_PATH = "./lua/src"
-c_compiler.CFLAGS_BASE = "--std=c99 -Wall -g"
+c_compiler.CFLAGS_BASE = "--std=c99 -Wall -g -Wno-unused-function"
```
gligneul: `c_compiler.CFLAGS_WARNINGS` would be nice.

titan-compiler/coder.lua
```diff
@@ -297,6 +382,22 @@ local function check_tag(typ, slot)
...
+    elseif tag == types.T.Integer  then return "LUA_TNUMINT"
+    elseif tag == types.T.Float    then return "LUA_TNUMFLT"
+    elseif tag == types.T.String   then return "LUA_TSTRING"
+    elseif tag == types.T.Function then return "LUA_TFUNCTION"
+    elseif tag == types.T.Array    then return "LUA_TTABLE"
+    elseif tag == types.T.Record   then error("not implemented")
+    else error("impossible")
+    end
+    return util.render(tmpl, {SLOT = slot})
+
```
gligneul: s/\n//

titan-compiler/coder.lua
```diff
@@ -1008,18 +1104,15 @@ generate_stat = function(stat, ctx)
                 stat.exp._type, var_lvalue.slot_address, exp_cvalue,
                 var_lvalue.parent_pointer)
             assign_stat = util.render([[
-                if (isempty(${SLOT})) {
-                    luaL_error(L,
-                        "out of bounds (inside array part) at line %d",
-                        ${LINE});
-                }
+                if (!${CHECK_TAG}) { titan_runtime_array_type_error( L, ${LINE}, ${EXPECTED_TAG}, ${SLOT} ); }
```
gligneul: Could we break the line here to maintain a pattern?
Also, no need for the extra spaces between the parenthesis.

titan-compiler/coder.lua
```diff
@@ -1134,11 +1227,7 @@ generate_var = function(var, ctx)
             ${T_CSTATS}
             ${K_CSTATS}
             ${UI_DECL} = ((lua_Unsigned)${K_CVALUE}) - 1;
-            if (${UI_NAME} >= ${T_CVALUE}->sizearray) {
-                luaL_error(L,
-                    "out of bounds (outside array part) at line %d",
-                    ${LINE});
-            }
+            if (${UI_NAME} >= ${T_CVALUE}->sizearray) { titan_runtime_array_bounds_error(L, ${LINE}); }
```
gligneul, line 1218: See above.

titan-compiler/coder.lua
```diff
@@ -1452,33 +1539,41 @@ generate_exp = function(exp, ctx)
...
-                }
-                if (!${CHECK_TAG}) {
-                    luaL_error(L,
-                        "wrong type for array element at line %d, "
-                        "expected %s but found %s",
-                        ${LINE}, ${EXP_TYPE}, lua_typename(L, ttype(${SLOT})));
-                }
+            local cstats = util.render([[
+                ${EXP_STATS}
+                if (!${CHECK_TAG}) { titan_runtime_array_type_error( L, ${LINE}, ${EXPECTED_TAG}, ${SLOT} ); }
```
gligneul: See above.

titan-compiler/coder.lua
```diff
@@ -1452,33 +1539,41 @@ generate_exp = function(exp, ctx)
...
+            local cvar = ctx:new_cvar(ctype(exp._type))
+            local cstats = util.render([[
+                ${EXP_STATS}
+                ${VAR_DECL} = ${GET_SLOT};
+            ]], {
+                EXP_STATS = exp_cstats,
+                VAR_DECL = c_declaration(cvar),
+                GET_SLOT = get_slot(exp.var._type, lvalue.slot_address),
+            })
+            return cstats, cvar.name
```
gligneul, line 1541: Since we are returning here we could remove the return at line 1580.

gligneul (changes requested):

titan-compiler/coder.lua
hugomg: I put it on a single line because I found it much easier to work with the generated C code when it was that way. The group of statements behind an array access became more visually distinct and it was much easier to comment the check in and out...

titan-compiler/coder.lua
hugomg: I wish this comment had been attached to line 1580 :)

gligneul: GIthub wouldn't allow me.

titan-compiler/coder.lua
gligneul: Could you fix the spaces between parenthesis here too.

titan-compiler/coder.lua
gligneul: Could we change the other error checks to be one liners too?

titan-compiler/coder.lua
```diff
@@ -518,10 +613,7 @@ local function generate_lua_entry_point(tl_node)
     local check_nargs = util.render([[
         ${NARGS_DECL} = cast_int(L->top - (${BASE} + 1));
         if (${NARGS} != ${EXPECTED}) {
-            luaL_error(L,
-                "wrong number of arguments to function, "
-                "expected %d but received %d",
-                ${EXPECTED}, ${NARGS});
+            titan_runtime_arity_error(L, ${EXPECTED}, ${NARGS});
```
gligneul: Could we change the other error checks to be one liners too?

titan-compiler/coder.lua
gligneul: Could you fix the spaces between parenthesis here too?

gligneul (changes requested):

titan-compiler/coder.lua
```diff
@@ -545,11 +635,9 @@ local function generate_lua_entry_point(tl_node)
         table.insert(check_types, util.render([[
             ${SLOT_DECL} = ${SLOT_ADDRESS};
             if (!${CHECK_TAG}) {
-                luaL_error(L,
-                    "wrong type for argument %s at line %d, "
-                    "expected %s but found %s",
-                    ${PARAM_NAME}, ${LINE}, ${EXP_TYPE},
-                    lua_typename(L, ttype(${SLOT_NAME})));
+                titan_runtime_argument_type_error(
```
gligneul: Also make this one liner

titan-compiler/coder.lua
```diff
@@ -1420,11 +1501,9 @@ generate_exp = function(exp, ctx)
                 table.insert(body, util.render([[
                     ${SLOT_DECL} = s2v(L->top-1);
                     if (!${CHECK_TAG}) {
-                        luaL_error(L,
-                            "wrong type for function result at line %d, "
-                            "expected %s but found %s",
-                            ${LINE}, ${EXP_TYPE},
-                            lua_typename(L, ttype(${SLOT})));
+                        titan_runtime_function_return_error(
```
gligneul: Also make this one liner

gligneul (changes requested):

gligneul (approved):

PR #200, opened 2018-04-12
Titan runtime

hugomg:


Makefile
```diff
@@ -0,0 +1,16 @@
+CPPFLAGS:=-I./lua/src
+CFLAGS:=--std=c99 -g -Wall -O2 -fPIC -shared
```
gligneul: Clang gives me this warning
```
clang: warning: argument unused during compilation: '-shared' [-Wunused-command-line-argument]
```

Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+.PHONY: all lua clean
+
+all: lua runtime/tcore.o
```
gligneul: I think we should add another phony called runtime (or titanruntime) and add `runtime/tcore.o` as a dependency of it.

Makefile
```diff
@@ -0,0 +1,16 @@
...
+
+lua:
+	cd lua/src; make linux-readline
+
+runtime/tcore.o: runtime/tcore.c runtime/tcore.h
+	$(CC) $(CPPFLAGS) $(CFLAGS) -c $< -o $@
```
gligneul: I think this is the default rule already, we shoudn't need to specify it.

Makefile
```diff
@@ -0,0 +1,16 @@
+CPPFLAGS:=-I./lua/src
+CFLAGS:=--std=c99 -g -Wall -O2 -fPIC -shared
+
+.PHONY: all lua clean
+
+all: lua runtime/tcore.o
+
+lua:
+	cd lua/src; make linux-readline
```
gligneul: Why not use the lua root makefile (instead of the one at src dir)?

gligneul: The linux-makefile target won't work in macos. 😢 
We could have a TARGET variable (and an error if it isn't defined).

runtime/tcore.c
```diff
@@ -0,0 +1,114 @@
+#include "tcore.h"
+
+#include "lua.h"
+#include "lauxlib.h"
+
+#include "lobject.h"
+#include "lstate.h"
+#include "ltm.h"
+
+const char * titan_tag_name(int raw_tag)
```
gligneul: Since this isn't a generated file anymore I think we should be strict about code style.
`const char *titan_tag_name(int raw_tag)`

runtime/tcore.c
```diff
@@ -0,0 +1,114 @@
...
+    } else {
+        return ttypename(novariant(raw_tag));
+    }
+}
+
+void titan_runtime_arity_error(
+    lua_State *L,
+    int expected,
+    int received
+){
```
gligneul, line 25: Could we close this `)` at the previous line (if yes fix all functions).

runtime/tcore.h
```diff
@@ -0,0 +1,32 @@
+#ifndef TCORE_H
+#define TCORE_H
+
+#include "lapi.h"
+
+#define TITAN_NORETURN __attribute__((noreturn))
+#define TITAN_UNREACHABLE __builtin_unreachable()
+
+const char * titan_tag_name(int raw_tag);
```
gligneul: `const char *titan_tag_name(int raw_tag);`

gligneul (changes requested):

Makefile
hugomg: The Lua root makefile is currently bugged and is missing the "linux-readline" option. This has been reported to Roberto and will be fixed in the work2 release

runtime/tcore.c
hugomg: The pretty printer prefers the `)` on a line by itself, similar to how `}` works. I would like to be consistent with that.

Makefile
```diff
@@ -0,0 +1,28 @@
...
+
+default:
+	@echo "Please do 'make PLATFORM' where PLATFORM is one of these:"
+	@echo "  linux-readline linux-noreadline macos"
+
+clean:
+	cd lua/src && $(MAKE) clean
+	rm -f ./runtime/*.o ./runtime/*.a
+
+
```
gligneul: s/\n//

Makefile
```diff
@@ -0,0 +1,28 @@
...
+linux-readline:   runtime/tcore.o lua-linux-readline
```
gligneul: I think we have a runtime phony target, so it will be easier to add another files later.

```
linux-readline:   runtimelib lua-linux-readline 

runtimelib: runtime/tcore.o
```

gligneul (changes requested):

Makefile
hugomg: I would prefer adding this when we add those extra files (YAGNI). Unless we can find a way to package all the ".o" into a single file to link against (which is what I was trying to do with the .a)

gligneul (approved):

PR #206, opened 2018-04-14
Rehash

hugomg:


mascarenhas:
Can we get a high-level overview on what this PR is doing, specially given that it is messing with `ltable.h`?

hugomg:
This PR fixes a rare corner case where Lua sequences might not be fully stored inside the array part of the table. For example, with Lua's current implementation of tables, the following code produces a table where elements 1-8 are stored in the array part and 9-24 are stored in the hash part:

    local xs = {}
    for i = 1,5 do
        xs[i] = i
    end
    for i = 24,9,-1 do
        xs[i] = i
    end
    for i = 6,8 do
       xs[i] = i
    end

If we naively assume that all indices in this sequence are in the array part, trying to read or write to `xs[24]` in Titan will produce an out-of-bounds error. To fix this, whenever we try to read or write to an index outside the array part, we first attempt to "renormalize" the table and put everything that should be in the array part inside the array part and only if the index is still outside the array part then we produce an out of bounds error.

This change has a big performance implication in benchmarks with lots of array reads and writes such as matmulsum, because now that the "out of bounds" function is no longer "noreturn" (it might return successfully when trying to read from a valid index in a denormalized table) the C compiler generates less optimal code for the inner loop. However, we should be able to work around this by lifting the bounds check out of the loop in an optimization pass (I verified this by compiling some "artisanal titan" test cases).

Another change in this PR is that it goes back to the old behavior of table reads. Now they are allowed to write to the first free position at the end of the array so growing a table with `t[#t+1] = x` now works again. However, we still only allow to remove table elements with table.remove. This makes the code a bit faster by getting rid of a type check when performing an array write and by allowing people to avoid `#t` with `for i = 1, N do t[i] = x end`. This implementation isn't very strict though, and also allows writing to any index inside "sizearray". For example, `t[#t + 2]` will also succeed most of the time. 

One smelly part of the patch is that it adds a variant of the rehash function inside ltable.c. We can't use rehash directly because it assumes that it is being called when adding a new key to the table but in titan we can also call rehash when performing an array read. A less invasive approach might be to implement this rehash function inside the titan runtime but to do that we would need to copy all the static functions that rehash depends on as well. I think in the long term this might be the only option but for now I thought it was a better idea to put the new rehash function inside ltable.c, in case we change it to something else or perhaps even remove the need for it entirely.

mascarenhas:
Ok, I was not aware that Titan arrays had been redefined from possibly sparse maps of integer to some type to dense (maybe with no holes?) sequences in this branch. 

Given that we already had some benchmarks getting pretty close to code using C arrays even with regular Lua sparse arrays when we took care not to call `#` in a loop, I wonder what is the real impact of this will be.

hugomg:
That is a good point. The lablua branch had inadvertently changed the meaning to forbid (not always allow) arrays/tables with holes because I had falsely assumed that this is what we had always been doing. I am going to check what would happen if we went back to the original behavior from master. It certainly would make the code simpler (by getting rid of the need for rehash and things like that) but I will need to run some new experiments to study the performance impact.

hugomg:
Closing this for now because it will need extensive rethinking.

PR #212, opened 2018-04-15
Make generate_exp always return a single token

hugomg:
That is, a variable name or constant literal.

Instead of compiling `(a + b) + c`  to `(a + b) + c)` the coder will now output something like

    tmp1 = a + b
    tmp2 = tmp1 + c

This should make the order of evaluation wholly unambiguous and ensure that the output of generate_exp can be safely passed to macros or macro-like things (like our version of luaV_div and so on).

### DIFF
--- a/runtime/tcore.c
+++ b/runtime/tcore.c
@@ -119,3 +119,10 @@ void titan_runtime_divide_by_zero_error(
     luaL_error(L, "attempt to divide by zero");
     TITAN_UNREACHABLE;
 }
+
+void titan_runtime_mod_by_zero_error(
+    lua_State *L
+){
+    luaL_error(L, "attempt to perform 'n%%0'");
+    TITAN_UNREACHABLE;
+}
--- a/runtime/tcore.c
+++ b/runtime/tcore.c
@@ -112,3 +112,10 @@ void titan_runtime_function_return_error(
     );
     TITAN_UNREACHABLE;
 }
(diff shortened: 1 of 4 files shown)


runtime/tcore.c
```diff
@@ -112,3 +112,17 @@ void titan_runtime_function_return_error(
     );
     TITAN_UNREACHABLE;
 }
+
+void titan_runtime_divide_by_zero_error(
+    lua_State *L
+){
+    luaL_error(L, "attempt to divide by zero");
```
gligneul: We should the offending line to the error message.

runtime/tcore.c
```diff
@@ -112,3 +112,17 @@ void titan_runtime_function_return_error(
...
+    TITAN_UNREACHABLE;
+}
+
+void titan_runtime_mod_by_zero_error(
+    lua_State *L
+){
+    luaL_error(L, "attempt to perform 'n%%0'");
```
gligneul: We should the offending line to the error message.

gligneul (dismissed):

runtime/tcore.c
hugomg: I think we could add the line numbers here to be consistent with how the rest of the code currently works. However, I wonder if there would be a way to get rid of all these parameters and get the call stack info from elsewhere.

PR #214, opened 2018-04-16
Make Titan bitshift operators behave the same as Lua

hugomg:
In Lua, shifting by negative numbers shifts in the opposite direction and
shifting by an amount greater than or equal to the width of the integer type
results in 0.

Previously we were compiling titan bitshifts directly into C bitshifts, which
would run into undefined behavior in those cases.

### DIFF
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -541,7 +541,7 @@ describe("Titan coder", function()
         end)
     end)
 
-    describe("Two-s compliment integer arithmetic", function()
+    describe("Lua vs C operator semantics", function()
         it("unary (-)", function()
             run_coder([[
                 function f(x:integer): integer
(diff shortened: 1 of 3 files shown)


spec/coder_spec.lua
```diff
@@ -583,13 +583,28 @@ describe("Titan coder", function()
...
-                assert(-1 == test.f(-10, -3))
+                assert( 10 %  3 == test.f( 10,  3))
+                assert( 10 % -3 == test.f( 10, -3))
+                assert(-10 %  3 == test.f(-10,  3))
+                assert(-10 % -3 == test.f(-10, -3))
                 assert(0 == test.f(math.mininteger, -1))
             ]])
         end)
+
+        it("(>>)", function()
```
gligneul, line 594: Do we have tests for `<<`?

gligneul (changes requested):

gligneul (approved):

PR #222, opened 2018-04-25
Test corner cases for all operators

hugomg:


test_operators.lua
```diff
@@ -0,0 +1,61 @@
...
+            local ok, err = check(f_lua, f_titan, {x, y})
+            if not ok then
+                error(string.format("%s %s %s: %s",
+                    tostring(x), op_str, tostring(y), err))
+            end
+        end
+    end
+end
+
+return {
```
gligneul: I think we should use the protocol we are already using in other modules:
```
local module = {}

function module.foo()
end

return module

File: test_operators.lua [hunk 2]
```diff
@@ -0,0 +1,61 @@
...
+end
+
+local function are_same(a, b)
+    return (a == b) or (isnan(a) and isnan(b))
+end
+
+local function check(f_lua, f_titan, args)
+    local ok1, a = pcall(f_lua, table.unpack(args))
+    local ok2, b = pcall(f_titan, table.unpack(args))
+    if ok1 ~= ok2 then
```
gligneul: If both calls crash this will be false and the values will be compared bellow, is this expected?
Shouldn't we do check something like:
```
if not ok1 or not ok2 then
```

test_operators.lua
```diff
@@ -0,0 +1,61 @@
...
+local function isnan(x)
+    return x ~= x
+end
+
+local function are_same(a, b)
+    return (a == b) or (isnan(a) and isnan(b))
+end
+
+local function check(f_lua, f_titan, args)
+    local ok1, a = pcall(f_lua, table.unpack(args))
```
gligneul: ok1 and ok2 are not good variable names, we could be a bit more specific like ok_lua and ok_titan.

test_operators.lua
```diff
@@ -0,0 +1,61 @@
...
+            (ok2 and "didn't crash" or "crashed"))
+    end
+    if ok1 and ok2 and not are_same(a, b) then
+        return false, string.format("(lua: %s, titan: %s)",
+            tostring(a), tostring(b))
+    end
+    return true
+end
+
+local function check_unop(op_str, f_lua, f_titan, typ1)
```
gligneul: Why typ1 instead of typ? Seems a typo but I'm not sure.

test_operators.lua
```diff
@@ -0,0 +1,61 @@
...
+    for _, x in ipairs(values[typ1]) do
+        local ok, err = check(f_lua, f_titan, {x})
+        if not ok then
+            error(string.format("%s %s: %s",
+                op_str, tostring(x), err))
+        end
+    end
+end
+
+local function check_binop(op_str, f_lua, f_titan, typ1, typ2)
```
gligneul: Again, we could replace 1 and 2 with better names. Suggestion:
```
s/typ1/lhs_type/
s/typ2/rhs_type/
```

test_operators.lua
```diff
@@ -0,0 +1,61 @@
...
+
+local function isnan(x)
+    return x ~= x
+end
+
+local function are_same(a, b)
+    return (a == b) or (isnan(a) and isnan(b))
+end
+
+local function check(f_lua, f_titan, args)
```
gligneul: Why manually implement vararg instead of using `...`?

test_operators.lua
gligneul: `a` and `b` are also bad names, why not `ret_lua` (or `*_lua`)?

test_operators.lua
```diff
@@ -0,0 +1,61 @@
+-- Used in checker_spec
+
+local values = {
+    ["boolean"] = { true, false },
+    ["integer"] = {
+        math.mininteger, -10, -3, -2, -1, 0, 1, 2, 3, 10, math.mininteger,
```
gligneul: Is testing -3 and 3 really useful?

test_operators.lua
```diff
@@ -0,0 +1,61 @@
...
+    },
+    ["float"] = {
+        -math.huge, -2.0, -1.0, 0.0, 1.0, math.pi
```
gligneul: We could test +math.huge also?

spec/coder_spec.lua
```diff
@@ -284,7 +284,7 @@ describe("Titan coder /", function()
             end
         ]]))
 
-        it("local variables", function()
+               it("local variables", function()
```
gligneul: Indentation is wrong here.

spec/coder_spec.lua
```diff
@@ -298,315 +298,126 @@ describe("Titan coder /", function()
...
-                run_binop_test("lshift", "<<", "0xdead", "-1")
-                run_binop_test("lshift", "<<", "0xdead", " 100")
-                run_binop_test("lshift", "<<", "0xdead", "-100")
-                run_binop_test("lshift", "<<", "0xdead", "math.maxinteger")
-                run_binop_test("lshift", "<<", "0xdead", "math.mininteger")
-            end)
-        end)
+        setup_unop("neg_int", "-",   "integer", "integer")
+        setup_unop("bnot",    "~",   "integer", "integer")
+        setup_unop("not_bool","not", "boolean", "boolean")
```
gligneul: Missing a space after the first comma.

spec/coder_spec.lua
```diff
@@ -298,315 +298,126 @@ describe("Titan coder /", function()
...
+
+        setup_binop("add_int"       , "+", "integer", "integer", "integer")
```
gligneul: In the previous block the comma was aligned with argument in the left, here it is aligned in the right. We should be consistent.
Also, we could align the other arguments.

gligneul (changes requested):

gligneul:
Github is showing some of my comments as outdated because I reviewed them commit by commit, please take a look in all of them.

gligneul (approved):

PR #223, opened 2018-04-26
Records improvements

gligneul:


titan-compiler/coder.lua
```diff
@@ -476,34 +476,60 @@ end
...
-            index = index + 1
+            pos[field.name] = gc_index
+            gc_index = gc_index + 1
+        else
+            pos[field.name] = prim_index
+            prim_index = prim_index + 1
         end
     end
-    return { pos = pos, size = index }
+    return { pos = pos, gc_size = gc_index, prim_size = prim_index }
```
hugomg: Could we be more defensive and have separate `gc_pos` and `prim_pos` tables instead of a shared `pos` table? This way, if there is a compiler bug and we try to use a prim field as a gc field or vice versa we get a compiler crash instead of it silently doing the wrong thing.

titan-compiler/coder.lua
```diff
@@ -476,34 +476,60 @@ end
...
+    local layout = rec._layout
+    if layout.prim_size == 0 then
+        return 0
+    else
+        return string.format("sizeof(%s)", rec_struct_name(rec))
+    end
+end
+
 local function rec_declare_struct(rec)
+    if rec._layout.prim_size == 0 then return "" end
```
hugomg: Please add a comment clarifying that empty structs are not allowed by the C standard and that they are a GCC extension. (I think here is the best spot but anywhere would be fine)

titan-compiler/coder.lua
```diff
@@ -476,34 +476,60 @@ end
...
+
     local fields = {}
     for _, field in ipairs(rec.field_decls) do
         local typ = rec._field_types[field.name]
         if not types.is_gc(typ) then
-            local decl = string.format("%s %s;", ctype(typ), field.name)
+            local name = rec_field_name(rec, field.name)
+            local field = new_cvar(name, ctype(typ))
```
hugomg: We could put the original Titan field name in a comment:

    new_cvar(name, ctype(typ), field.name)

hugomg (changes requested):

titan-compiler/coder.lua
gligneul: Will do.

titan-compiler/coder.lua
gligneul: Will do.

titan-compiler/coder.lua
gligneul: Will do.

hugomg (approved):

PR #230, opened 2018-05-13
Support for compiling stand-alone executables

hishamhm:
This PR adds support for compiling stand-alone executables.

If a module contains a `main` function with signature `function({string}):integer`,
that is the main entry point of the program. The input argument is an array
of strings containing the command-line arguments, and the integer return value
is the exit code of the program.

Only one of the modules given to `titanc` may contain a `main` function.

When compiling a program as a single executable, `titanc` does not use dynamic linking to resolve references in imports. Instead, it generates simple `extern` declarations that are resolved in the linking step.

The `examples.sdl_demo` example is modified to demonstrate this, moving the helper function into a separate module.

Additionally, this adds a `-v`/`--verbose` flag to `titanc`, which makes it display the C compiler invocations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,9 +45,16 @@ binary path, a semicolon-separated list of paths
 or `TITAN_PATH` environment variable). A module gets compiled if its `.titan` file
 is newer than its binary, or a binary does not exist.
 
-If everything is all right with your modules this will generate shared libraries
-(in the same path as the module source) that you can `require` from Lua, and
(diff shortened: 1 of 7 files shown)


mascarenhas (approved):
It is great to finally have this! Later we can have a way for Titan to run Lua code, and make sure that this code can access the modules that the Titan application has, but this is a great start.

PR #231, opened 2018-05-15
Changes to build and run Titan modules and applications with stock Lua (without changes needed to `luaconf.h`)

mascarenhas:
Link Titan modules statically with part of the Lua object files (only change over a regular Lua installation is that they need to be compiled with `-fPIC`, but no changes to `luaconf.h` are needed.

All `static` functions that every Titan module/app uses should be in a separate `titan.c` file (some of these functions may be compiled from Lua and specialized, with proper attribution).

Titan applications do not need Lua object files compiled with `-fPIC`, for now we can just have one set of object files, in the future we might want to have two.

I am also trying to get Travis to do OSX builds of Titan, but as I do not have a local OSX machine myself I will leave this out of this PR if it turns out to be too much trouble.


hishamhm:
> I am also trying to get Travis to do OSX builds of Titan, but as I do not have a local OSX machine myself I will leave this out of this PR if it turns out to be too much trouble.

I'm going to push a commit here attempting to fix this, I checked the changes against the LuaRocks `.travis.yml`.

mascarenhas:
Right now it seems like the Travis OSX worker is busted, it is erroring while downloading Python.

mascarenhas:
Something pretty odd just happened with this branch, it is showing the latest commits that were already merged into `master`.

hishamhm:
@mascarenhas my bad, just refresh :) I messed it up for a second when force-pushing a fixed-fix.

mascarenhas:
We can already remove the `lua` folder from the repository, by the way. About this PR, I still have not tested if Titan executables are building correctly (I suspect not), and after fixing this we can merge and do the specialization of `getgeneric` some other time.

mascarenhas:
Looks like the FFI needs some fixes to work on OSX. We should just leave the Travis build commented out for now and try to do these fixes in  another PR, preferably with access to an OSX machine.

hishamhm:
All right, macOS ran, but c-parser is failing to parse some standard library headers there (those are usually the ones that push the parser the hardest).

hishamhm:
@mascarenhas agreed!

titan-runtime/titan.h
```diff
@@ -0,0 +1,58 @@
...
+    size_t len;
+    static char _cvtbuff[MAXNUMBER2STR];
+    len = lua_number2str(_cvtbuff, sizeof(_cvtbuff), f);
+    return luaS_newlstr(L, _cvtbuff, len);
+}
+
+LUAI_FUNC const TValue* getgeneric(Table* t, const TValue* key);
+LUAI_FUNC void *loadlib (lua_State *L, const char *file);
+LUAI_FUNC void *loadsym (lua_State *L, void *lib, const char *sym);
+LUAI_FUNC const TValue *getgeneric (Table *t, const TValue *key);
```
hishamhm: getgeneric is duplicated here

mascarenhas: oops, pushing a fix

mascarenhas:
Ok, will make sure we can still compile Titan applications, and not just Lua modules, later.

mascarenhas:
All systems go @hishamhm!

hishamhm (approved):
@mascarenhas last missing bit: could you update the build/install instructions in README.md? I'll leave the PR approved so you can merge once that is in!

PR #232, opened 2018-05-16
Builtin functions

mascarenhas:
Adds the capability of having builtin functions in the Titan runtime that are callable directly from Titan code, and defines a few builtins available to all Titan modules (such as `print`).

The idea is to use builtin functions for standard library functions that need to access Titan's Lua state, and for functions that do not need to access the Lua state but where it is incovenient or impossible to define the function by a combination of Titan code plus FFI calls.

These builtin functions have the Titan calling convention, apart from the addition of variadic arguments, encoded as an `int` parameter that takes the size of the variadic argument list, and a standard C variadic parameter.

The C name of a builtin function is always `titan_<name>` for the builtins that are not part of any module, such as `print` and `tostring`, and `titan_<mangled_fqtn>` for builtins declared in as part of a module, where `mangled_fqtn` is the fully-qualified name of the function with all dashes and dots replaced by underscores.

### DIFF
--- a/doc/manual.md
+++ b/doc/manual.md
@@ -400,6 +400,58 @@ Casts from `float` to `integer` fail if it is not an integral value, and if
 this value is outside the allowable range for integers. Casts from `value`
 fail if the value does not have the target type, or cannot be converted to it.
 
+## The standard library
+
+### Basic functions
+
+You can call these functions directly from anywhere in a Titan module,
+without needing to import anything.
+
+#### assert(v: value, msg: string): string
+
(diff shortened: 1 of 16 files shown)


mascarenhas:
By the way, let's not get hung up on the specific interfaces and implementations for the builtin functions included in this PR, the lack of option types (and optional parameters) and proper error handling in Titan right now restricts what we can do. These builtin functions can improve later as the type system improves.

mascarenhas:
Rebased this onto the current master.

PR #233, opened 2018-05-17
No C global state in modules

mascarenhas:
This PR removes most of the C global state in Titan modules, the only thing that is left is a cache for tags. We are passing all the tests that we currently have, but I still need to write tests for functionality that this PR enables, such as loading the same Titan module from two distinct Lua states (their module variables will be local to the state, as it should be), or linking module B inside module A.so, while having a separate B.so, and having the B that A imports and the other B require'd by Lua refer to the same module instance.

The canonical module instances for each Lua state, as well the canonical tags and metatables for record types, are kept in the Lua registry, but we only need to hit the registry at module initialization, no accesses to the registry are done on normal execution of Titan code.

One thing that we lose is the ability to leverage GCC's constant propagation to propagate local module variables that have numeric values that can be know at compile-time, but we can add a constant propagation AST->AST pass later on.

### DIFF
--- a/titan-compiler/coder.lua
+++ b/titan-compiler/coder.lua
@@ -2981,24 +2981,6 @@ function coder.generate(modname, ast, is_dynamic)
         }))
     end
 
-    local nslots = initctx.nslots
-
-    table.insert(initvars, 1, render([[
-        /* reserve needed stack space */
-        if (L->stack_last - L->top > $NSLOTS) {
-            if (L->ci->top < L->top + $NSLOTS) L->ci->top = L->top + $NSLOTS;
-        } else {
-            lua_checkstack(L, $NSLOTS);
-        }
-        TValue *_base = L->top;
-        L->top += $NSLOTS;
-        for(TValue *_s = L->top - 1; _base <= _s; _s--) {
-            setnilvalue(_s);
-        }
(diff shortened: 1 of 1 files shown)


mascarenhas:
One thing that I also did in this PR was to make the code that runs only once in module initialization to be as well-behaved as possible in terms of using the Lua API instead of the internals, which hopefully will make decrease the surface area for changes between one Lua version to the other.

hishamhm:
I am going to try to rebase this PR on top of `master` without the `primitives`
branch, so this can be merged without being held back by the discussion
that is going to be necessary there.

mascarenhas:
> I am going to try to rebase this PR on top of `master` without the `primitives` branch, so this can be merged without being held back by the discussion that is going to be necessary there.

The easiest way to this is probably to do a new commit to this branch that removes the foreign functions, then squashing the the commits from both branches together and applying then on master...

But the objection was to foreign functions exposed to the programmer, not to the concept itself, so I can just not expose then for now. If I do this on the primitives branch we can probably merge both branches cleanly without losing history.

mascarenhas:
Ok, now this is a dead-end branch, closing this PR and killing the branch. Thanks for all the fish!

PR #234, opened 2018-05-17
tests: module local variables in application

hishamhm:
I opened a PR with a failing test case instead of filing a bug. :)


mascarenhas:
If the module was linked statically its initialization function was not being called, fixed! Had already fixed this in the 'no-global-state' PR, also.

mascarenhas (approved):

mascarenhas:
I approved this, but as I said, this is subsumed by another PR.

PR #235, opened 2018-05-18
[master] remove troublesome global C state from modules

hishamhm:
This is a rebase into `master` of PR #233.

### DIFF
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -100,6 +100,10 @@ local function run_coder_app(titan_code, main, estatus, eout)
 end
 
 describe("Titan code generator", function()
+    setup(function ()
+        os.remove("titan-runtime/titan.o")
+    end)
+
     after_each(function ()
         os.execute("rm -f *.o")
         os.execute("rm -f *.so")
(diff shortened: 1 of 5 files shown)


hishamhm:
@mascarenhas this made it much easier to follow! (and made me note that there was an update done to `codearray` that wasn't done in `codemap`; I updated and squashed it there — I'll probably refactor the common parts of these functions later)

hishamhm:
I would prefer to merge this one instead (and work using this version over the weekend, until we get a chance to discuss primitives in the next meeting)

mascarenhas:
I am fixing a couple of loose ends and will push a commit soon, and then we can merge.

mascarenhas:
My commit addresses two issues:

* There was a bugfix for comparing two `value`s in the `primitives` branch that we should keep
* The slot indexes of a module's members must be deterministic, otherwise we cannot have separate compilation with the current scheme (we need a test for that)

mascarenhas:
Feel free to merge this into `master`!

mascarenhas (approved):

PR #236, opened 2018-05-21
test: call a method from a returned record

hishamhm:
A failing test case

### DIFF
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -1627,6 +1627,35 @@ describe("Titan code generator", function()
         assert.truthy(ok, err)
     end)
 
+    it("calls method from a returned record", function ()
+        local modules = {
+            foo = [[
+                record Point
+                  x: float
+                  y: float
+                end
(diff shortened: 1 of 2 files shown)


mascarenhas:
Fixed. The test for whether a function call is a foreign call was expecting the function expression to be an lvalue, but this is not always the case for method calls. I commited quick fix, but we should really improve the handling of all the different kinds of calls and receivers of these calls...

I am always a proponent of different types for different things, so we should probably reserve `Type.Function` for first-class functions and then have `Type.FirstOrderFunction`, `Type.ForeignFunction`, `Type.RecordFunction` (the current `Type.StaticMethod`), and `Type.Method` as distinct types for the kinds of callables that we have. It also wouldn't hurt to split `ExpCall` into `ExpCall` and `ExpMethodCall`...

@hishamhm, I will leave this for you to merge this so you can see this comment.

mascarenhas (approved):

hishamhm:
Agreed! Thank you, @mascarenhas!

PR #237, opened 2018-05-22
Mix static and dynamic loading, make static default when source is present

mascarenhas:
This PR makes static linking the default when source code for an imported module is present, but still allows dynamic linking if there is no source code but a shared library for the imported module.

The `titanc` compiler now generates either one shared library for each module passed in its command line if there is no `main` function present. Each of these shared libraries statically embed all their dependencies where source code was found.

If there is a `main` function in any of the modules passed to `titanc` a single executable will be generated, statically embedding all of the modules passed in the command line plus their dependencies where source code was present.


hishamhm:
> statically embedding all of the modules passed in the command line plus their dependencies where source code was present.

This is, to use a current buzzword, very "opinionated" behavior. I don't mind it myself, currently (would need to see with actual use how good/annoying this is — but I like the principle of picking an option and trying to do the helpful thing by default), but theoretically speaking, how hard would it be to make it disable-able via a flag?

mascarenhas:
> how hard would it be to make it disable-able via a flag?

Very easy, we just need to tweak the default compile-time module loader to respect these flags. The hard work of mixing static and dynamic imports is already being done by the code generator, where you now pass the set of dependencies that are statically compiled instead of blanket flag that either disables dynamic loading or makes everything dynamically loaded.

mascarenhas:
To be honest l, I would not mind if the policy for `titanc` would be no dynamic loading of Titan modules from Titan code, while leaving the plumbing for dynamic loading in place, along with the tests for that. It will probably be a while until someone requests a way to distribute binary-only modules.

hishamhm (approved):
pushed a commit fixing a few doc typos, otherwise LGTM!

PR #239, opened 2018-05-29
Option types

mascarenhas:
The idea of this first version of option types (`t?`) is to be "gradually typed", that is, Titan has a distinction between `t` and `t?`, but will implicitly cast between those two types (of course, casting from `t?` to `t` will fail at runtime if the value is not there). The rationale is to not bother the programmer with having to prove that the value is always there, as that is very finnicky in an imperative language, but fail fast if he thinks he is handling a real value but the value is not actually there, so "nulls" do not propagate.

A key piece of this is to never infer an option type, the programmer has to explicitly "opt-in" to having an expression be nullable. So even though reading from an array always gives back an option type, *initializing* a variable from an array will always check if the value exists and type it accordingly, unless the programmer explicitly tells Titan he wants to propagate the option with something like `local x: integer? = arr[i]`.

Explicitly keeping a value optional is some thing that we expect to be common, so there is syntax to infer option types: `local x? = <exp>`.

Option types play well with the usual Lua idioms for `and` and `or`. The type of `x and y` is an option type constructed from the type of `y`, and the type of `x or y` is `t` if `x` has type `t?` and `y` has type  `t`, so the "ternary operator" `<cond> and <exptrue> or <expfalse>` works as expected.


hishamhm (approved):
LGTM!

PR #243, opened 2018-07-25
Recovery

sqmedeiros:


titan-compiler/typedecl.lua
```diff
@@ -9,7 +9,7 @@ return function(prefix, types)
             constructors[consname] = function(...)
                 local args = table.pack(...)
                 if args.n ~= #fields then
-                    error("missing arguments for " .. consname)
+                    error("missing arguments for " .. consname .. tostring(args.n) .. tostring(#fields))
```
hugomg, line 12: It seems that these were added for debugging but not remover later.
If we want to keep them they should have a space between the numbers at least.

titan-compiler/lexer.lua
```diff
@@ -180,7 +180,9 @@ for _, keyword in ipairs(keywords) do
 end
 
 lexer.NAME = Cmt(C(possiblename), function(_, pos, s)
+    --print('lexer', _, pos, s, is_keyword[s])
```
hugomg, line 183: These comments should be removed

titan-compiler/parser.lua
```diff
@@ -76,6 +96,9 @@ function defs.nil_exp(pos--[[, s ]])
 end
 
 function defs.number_exp(pos, n)
+    --print("number_exp", pos, n, math.type(n))
+    --assert(n ~= 52)
+    --assert(n ~= '52')
```
hugomg, line 101: More "debug" comments. Why is 52 important?

hugomg (changes requested):
Looking superficially, I noticed that there are some "debug comments" that were committed by accident. I didn't list all of them to avoid being repetitive.

I didn't review the meat of the parser yet. I'll need to learn more about how that works first.

What is the purpose of the annotatedTitan.txt and annotatedTitanMet.txt files? They look like they were generated automatically.

PR #244, opened 2018-08-04
fix header parser for OS X

mascarenhas:
OS X system headers have a bunch of custom attributes that do not mimic exactly the syntax of GCC attributes in the Linux headers, so make the parser more permissive to get around that (the parser ignores attributes anyway).

One of the OS X headers also uses `#include "..."` expecting the include file to be found in the same search path as the regular `#include <...>`, we work around that (probably not in the best way).

Finally, one of the OS X system headers has a function that takes a block, we now accept that and treat it just as a regular function pointer.

### DIFF
--- a/c-parser/c99.lua
+++ b/c-parser/c99.lua
@@ -64,7 +64,7 @@ defs["decl_ids"] = typed("string, number, table -> boolean, Decl?", function(_, 
     -- store typedef
     if elem(decl.spec, "typedef") then
         if not (decl.ids and decl.ids[1] and decl.ids[1].decl) then
-            return false, nil
+            return true
         end
         for _, id in ipairs(decl.ids) do
             local name = id.decl.name or id.decl.declarator.name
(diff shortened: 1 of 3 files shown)


hishamhm:
Great, thanks for digging this stuff! I adjusted the indentation level in some blocks and added a comment about the system-include location path in case we need to remember this in the future.

hishamhm (approved):

c-parser/c99.lua
```diff
@@ -64,7 +64,7 @@ defs["decl_ids"] = typed("string, number, table -> boolean, Decl?", function(_,
     -- store typedef
     if elem(decl.spec, "typedef") then
         if not (decl.ids and decl.ids[1] and decl.ids[1].decl) then
-            return false, nil
+            return true
```
hishamhm, line 67: oh, btw, do you remember why was this bit necessary?

mascarenhas: The OSX headers have some typedefs that end up being defined twice, and returning nil in a match-time capture (the `=>`) makes the pattern fail.

Issue #245, opened 2018-08-09
Titan Compiler doesn't run (titan-compiler.location module not found)


Jictyvoo:
# Problem

I installed Titan with luarocks for Lua 5.3
But everytime I want to tun compiler, with any file or without a file, appear a error message saying that:

## Error

/usr/bin/lua5.3: /usr/local/share/lua/5.3/titan-compiler/checker.lua:3: module 'titan-compiler.location' not found:No LuaRocks module found for titan-compiler.location
	no field package.preload['titan-compiler.location']
	no file '/usr/local/share/lua/5.3/titan-compiler/location.lua'
	no file '/usr/local/share/lua/5.3/titan-compiler/location/init.lua'
	no file '/usr/local/lib/lua/5.3/titan-compiler/location.lua'
	no file '/usr/local/lib/lua/5.3/titan-compiler/location/init.lua'
	no file '/usr/share/lua/5.3/titan-compiler/location.lua'
	no file '/usr/share/lua/5.3/titan-compiler/location/init.lua'
	no file './titan-compiler/location.lua'
	no file './titan-compiler/location/init.lua'
	no file '/home/jictyvoo/.luarocks/share/lua/5.3/titan-compiler/location.lua'
	no file '/home/jictyvoo/.luarocks/share/lua/5.3/titan-compiler/location/init.lua'
	no file '/usr/local/lib/lua/5.3/titan-compiler/location.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.3/titan-compiler/location.so'
	no file '/usr/lib/lua/5.3/titan-compiler/location.so'
	no file '/usr/local/lib/lua/5.3/loadall.so'
	no file './titan-compiler/location.so'
	no file '/home/jictyvoo/.luarocks/lib/lua/5.3/titan-compiler/location.so'
	no file '/usr/local/lib/lua/5.3/titan-compiler.so'
	no file '/usr/lib/x86_64-linux-gnu/lua/5.3/titan-compiler.so'
	no file '/usr/lib/lua/5.3/titan-compiler.so'
	no file '/usr/local/lib/lua/5.3/loadall.so'
	no file './titan-compiler.so'
	no file '/home/jictyvoo/.luarocks/lib/lua/5.3/titan-compiler.so'
stack traceback:
	[C]: in function 'require'
	/usr/local/share/lua/5.3/titan-compiler/checker.lua:3: in main chunk
	[C]: in function 'require'
[...]

PR #248, opened 2018-08-11
titanc: bail out of code generation on type errors

mascarenhas:
Fix for the odd behavior of `titanc` in Issue #245. Does not actually fix the underlying problem (missing `examples/sdl_quit.titan`), but will keep `titanc` from trying to compile files that have type errors (in this case a missing `import`).

### DIFF
--- a/titanc
+++ b/titanc
@@ -51,6 +51,8 @@ for _, modname in ipairs(args.module) do
     end
 end
 
+if #errors > 0 then exit(table.concat(errors, "\n")) end
+
 local libnames = {}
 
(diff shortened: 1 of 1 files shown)


hugomg (approved):
A line of code like this one should be somewhere but I don't know if it belongs in titanc or somewhere else, like the driver.

I'm accepting this PR but I'm not merging it in case you decide you still want to tweak it.

hishamhm:
I think it's fine in `titanc` for now as it does the "main pipeline" of the compilation

PR #255, opened 2018-09-23
fix some regressions with option types

mascarenhas:
Option types introduced some regressions, this PR fixes them and adds tests for those cases. 

First regression was indexing an array of arrays, the type checker was not coercing the option in the `e1` part of `e1[e2]` to its non-option version, so code like `x[...][...]` was failing because `x[...]` did not have an array type.

Second regression was comparing two option types, which should be like comparing two `value`s. Code like `a[1] == b[1]` that used to work now was having generating wrong C code because it was trying to compare two `TValue`s (the underlying C type of Lua slots and Titan option types) directly using `==`.

### DIFF
--- a/spec/coder_spec.lua
+++ b/spec/coder_spec.lua
@@ -2310,7 +2310,7 @@ describe("Titan code generator", function()
             ]])
         end)
 
-        it("compares two options", function ()
+        it("compares two values with option types", function ()
             run_coder([[
                 function fn(a: integer, b: integer): boolean
                   local t1 = { a }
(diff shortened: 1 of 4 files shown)


spec/coder_spec.lua
```diff
@@ -2310,6 +2310,18 @@ describe("Titan code generator", function()
             ]])
         end)
 
+        it("compares two options", function ()
```
hishamhm: "compares two values with optional types", maybe?

hishamhm (approved):
Just left a suggestion for a clearer test name

PR #257, opened 2018-11-23
add basic error handler function to validate approach

mascarenhas:
An experiment towards integrating Lua's error handling into Titan that makes errors thrown while executing Titan code throw an object that combines an error object with a stack trace that has both Titan and Lua functions.

The stack trace information for Titan functions has to be obtained through the C mechanisms (either the [backtrace](http://man7.org/linux/man-pages/man3/backtrace.3.html) function or [libbacktrace](https://github.com/ianlancetaylor/libbacktrace) or [libunwind](https://www.nongnu.org/libunwind/)), Titan functions have to be identified and demangled, and the Lua/Titan transitions identified and merged with the Lua stacktrace available through the `Callinfo` stack.


titan-runtime/titan.c
```diff
@@ -204,3 +205,13 @@ const TValue *getgeneric (Table *t, const TValue *key) {
     }
   }
 }
+
+int errorhandler(lua_State *L) {
+  void *buffer[255];
+  const int calls = backtrace(buffer, sizeof(buffer) / sizeof(void *));
```
hishamhm, line 211: We need some #ifdef around this to maintain portability. Even on Linux there are systems without `execinfo.h` (any non-glibc system, which includes most of embedded Linux and Alpine for small containers).

htop uses #ifdef HAVE_EXECINFO_H (which its `configure` script sets). Even if we don't put the actual detection for now, it's good to mark the knowingly-unportable code from the beginning so we don't have to hunt for it later. :)

hishamhm:
Nice! I think we could add some tiny test for this (run a failing program catching stderr and seeing we got _something_ resembling a backtrace in there?) and merge this initial version as is, then build on top of this in other PRs.

hishamhm:
(Funny this came along today, I was revisiting my endless "FFI refactor" branch during a flight just yesterday!)

titan-runtime/titan.c
mascarenhas: Oh, I am not intending to merge this as is, this was just a quick experiment, the next step as I said is to write the code to build the stack trace object, and not to print it. I just wanted to see if I could attach an error handler before the stack unwinds and if it seemed to have all necessary information available.

hishamhm: Let's merge as is! Small PRs mean easy reviews :)

hishamhm: Even if we don't do the full integrated stack trace having the two stacktraces is already a useful thing.

mascarenhas: I have amended the PR to not print the stack trace, but save it in the registry. There is a stub `titan.so` module with a single `backtrace()` function that reads this value. The backtrace that is saved is an amalgamation of the error message + Lua backtrace + C backtrace. Also added a test case.

hishamhm (approved):

